PR #3, opened 2018-12-07
Fix Grafana PV

thxCode:
Problem:
- Grafana can't mount PV correctly in Cloud
- Disabling monitoring will delete Grafana PV
- Redeploy Grafana with PV can't access after changing password

Solution:
- Use SecurityContext
- Use HELM resource-policy: https://github.com/helm/helm/blob/560b84d53c05885ba84384574520fdcf626937a3/docs/charts_tips_and_tricks.md#tell-tiller-not-to-delete-a-resource
- Remove `grafana-watch` and redesign Grafana provisioning

Issue:
https://github.com/rancher/rancher/issues/16953


thxCode:
@loganhz , @orangedeng

PR #4, opened 2018-12-07
Fix inaccessible Grafana

thxCode:
Problem:
Redeploying Grafana, which has changed password and owned PV, will not
be able to access again

Solution:
Remove Grafana account `Secret` and use Grafana provisioning instead of
`grafana-watch`

Issue:
https://github.com/rancher/rancher/issues/16962


thxCode:
@loganhz , @orangedeng

thxCode:
Close this, patch by another PR

PR #5, opened 2018-12-07
 Fix Grafana PV can't be mounted (Patch)

thxCode:
**Problem:**
Grafana still can't mount PV correctly in Cloud, because there are 2
`securityContext` and `fsGroup` only work on Pod spec.

**Solution:**
Change the `securityContext` to Pod spec.

**Issue:**
rancher/rancher#16953

### DIFF
--- a/charts/rancher-monitoring/v0.0.1/charts/grafana/templates/dashboards-configmap.yaml
+++ b/charts/rancher-monitoring/v0.0.1/charts/grafana/templates/dashboards-configmap.yaml
@@ -17,7 +17,7 @@ data:
       "access": "proxy",
       "basicAuth": false,
       "editable": false,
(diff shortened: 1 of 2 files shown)


charts/rancher-monitoring/v0.0.1/charts/grafana/templates/dashboards-configmap.yaml
```diff
@@ -17,7 +17,7 @@ data:
       "access": "proxy",
       "basicAuth": false,
       "editable": false,
-      "isDefault": true,
+      "isDefault:": true,
```
cjellick: this looks like youre putting the `:` back inside the quotes. do you want to do that?

cjellick:
LGTM

PR #12, opened 2019-02-13
Support editing Monitoring page

thxCode:
**Problem:**
- Cannot input like `x.y.z/k` label name into `nodeSelector`, `serviceSelectorLabels` fields
- ServiceAccount info will print after InitContainer running
- So many useless configuration items
- It's not easy to compare the `values.yaml` between sub-charts and root

**Solution:**
- Use array instead of object as values
- Remove the `cat` operation
- Remove them
- Only keep the different/private items in sub-charts, the public part save in root


loganhz:
Link to https://github.com/rancher/rancher/issues/17340

loganhz:
LGTM

PR #17, opened 2019-02-22
Should consist label in service monitor and logging charts

aiwantaozi:
Problem:
enable logging and moinitoring but can't see fluentd metric

Solution:
consist label and endpoint name

Issue:
https://github.com/rancher/rancher/issues/18327


thxCode:
LGTM, tested.

cjellick:
LGTM

PR #20, opened 2019-02-26
Add configurations for alibabacloud provider to external-dns

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/18324

### DIFF
--- a/charts/rancher-external-dns/v0.0.1/values.yaml
+++ b/charts/rancher-external-dns/v0.0.1/values.yaml
@@ -14,8 +14,15 @@ sources:
 # Allow external-dns to publish DNS records for ClusterIP services (optional)
 publishInternalServices: false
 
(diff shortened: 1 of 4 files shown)


charts/rancher-external-dns/v0.0.1/values.yaml
```diff
@@ -14,9 +14,16 @@ sources:
 # Allow external-dns to publish DNS records for ClusterIP services (optional)
 publishInternalServices: false
 
-## The DNS provider where the DNS records will be created (options: aws, google, inmemory, azure )
+## The DNS provider where the DNS records will be created (options: aws, google, inmemory, azure, alibaba )
```
thxCode: I think the name of Alibaba provider is `alibabacloud`.

gitlawr: Updated

thxCode:
LGTM

PR #21, opened 2019-02-27
upgrade fluentd image version

aiwantaozi:
problem:
before fluentd 1.3.1 version can't support add client cert for fluentd output

Solution:
upgrade fluentd to 1.3.3, but the related kafka gem also upgrade small
version, tested fluentd and kafka after upgrade version

Issue:
https://github.com/rancher/rancher/issues/18396


loganhz:
LGTM

PR #22, opened 2019-02-27
Use repository for image name key in templates

orangedeng:
In system-charts, we need to use the `repository` and `tag` to define
container's image name. After that, we can collect them together and
provide an images list we need for system charts.

Related issue: https://github.com/rancher/rancher/issues/18351


thxCode:
LGTM

PR #39, opened 2019-05-07
Enable Operator via operator.enabled or enabled

thxCode:


orangedeng:
LGTM

loganhz:
Issue: https://github.com/rancher/rancher/issues/20042

loganhz:
Cluster level monitoring, project level monitoring and monitoring operator are using the same charts.
But only monitoring operator should  deploy the operator.

The issue is now Cluster level monitoring also deploys operator.

 In cluster level monitoring,  we didn't define `operator.enable` and helm will always install operator.

If we change it to `condition: operator.enabled,enabled`, helm will check `operator.enabled` first and check `enabled` if `operator.enabled` is undefined. We have `enabled: false` in cluster monitoring app and `operator.enabled: false` in project level monitoring.

This is what we want, so existing 2.2.0 users can upgrade to 0.0.3 chart without upgrading rancher version.

alena1108:
@loganhz 

> This is what we want, so existing 2.2.0 users can upgrade to 0.0.3 chart without upgrading rancher version.

If we enable this functionality, we should call it out as an item to test in the ticket for QA

PR #40, opened 2019-05-08
The node exporter should be listening 0.0.0.0 instead of pod ip

orangedeng:
Fix the issue that node exporter crash when deploying
into the node without internal ip.

Related issue:
https://github.com/rancher/rancher/issues/19665


thxCode:
LGTM

PR #41, opened 2019-05-09
Add harbor chart v1.7.5-rancher1 for global registry

gitlawr:
Related issue: https://github.com/rancher/rancher/issues/1738

Notes:
1. global registry is available in HA setup.
2. Expose registry `/v2` endpoint to the same domain of Rancher server.
3. Expose `/api/projects` endpoint to Harbor core. It may be needed when the global registry is used as another Harbor's replication target. it will try to create a project via this endpoint when it does not exist.
4. Use `/registry` endpoint to aggregate Harbor UI and its API/services endpoints.
5. Add a proxy layer to help Harbor work with path prefix.


charts/harbor/v0.0.1/templates/registry/registry-cm.yaml
```diff
@@ -47,7 +47,7 @@ data:
         {{- if $storage.s3.encrypt }}
         encrypt: {{ $storage.s3.encrypt }}
         {{- end }}
-        {{- if $storage.s3.secure }}
+        {{- if kindIs "bool" $storage.s3.secure }}
```
thxCode: Why do we need `kindIs` here? When `$storage.s3.secure` is `false`, does it still return `true`?

charts/harbor/v0.0.1/templates/registry/registry-cm.yaml
```diff
@@ -120,7 +120,7 @@ data:
         {{- if $storage.oss.encrypt }}
         encrypt: {{ $storage.oss.encrypt }}
         {{- end }}
-        {{- if $storage.oss.secure }}
+        {{- if kindIs "bool" $storage.oss.secure }}
```
thxCode: Why do we need `kindIs` here? When `$storage.oss.secure` is `false`, does it still return `true`?

charts/harbor/v0.0.1/templates/proxy/deployment.yaml
```diff
@@ -0,0 +1,62 @@
...
+            - key: default.conf
+              path: default.conf
+      - name: filter-config
+        configMap:
+          name: "{{ template "harbor.proxy" . }}"
+          items:
+          - key: filters.conf
+            path: filters.conf
+    {{- with .Values.proxy.nodeSelector }}
+      nodeSelector:
```
thxCode: Make sure all workloads don't be scheduled into Windows host. You could choose to use `node selector` or `node affinity`.

charts/harbor/v0.0.1/templates/proxy/deployment.yaml
```diff
@@ -0,0 +1,62 @@
...
+{{ toYaml . | indent 8 }}
+    {{- end }}
+    {{- with .Values.proxy.affinity }}
+      affinity:
```
thxCode: Make sure all workloads don't be scheduled into Windows host. You could choose to use `node selector` or `node affinity`.

thxCode (commented):
I don't know `registry` components, I'm just concerned with some usage issues. Everything else looks good, I gave some suggestions and doubts.

charts/harbor/v0.0.1/templates/registry/registry-cm.yaml
gitlawr: Because `true` is the default value for the `secure` option. 
Originally, if the value is false, the if block is skipped and the false value is not used.
Now, if the value is false, it is bool kind so the if condition is true.
`false` and nil is indistinguishable in `if` statement so `kindIs` function is used here.

charts/harbor/v0.0.1/templates/registry/registry-cm.yaml
gitlawr: See above comment.

charts/harbor/v0.0.1/templates/registry/registry-cm.yaml
thxCode: This is not very intuitive to understand how to use `$storage.s3.secure`, could you add a new param to enter into this block?

charts/harbor/v0.0.1/templates/registry/registry-cm.yaml
thxCode: This is not very intuitive to understand how to use `$storage.oss.secure`, could you add a new param to enter into this block?

charts/harbor/v0.0.1/templates/registry/registry-cm.yaml
gitlawr: Updated

charts/harbor/v0.0.1/templates/proxy/deployment.yaml
gitlawr: Updated

thxCode:
LGTM

gitlawr:
@thxCode Added unit tests for the chart, PTAL.

thxCode:
LGTM

gitlawr:
Updated the chart version to `1.7.5-rancher1` per discussion

cjellick:
LGTM

PR #42, opened 2019-05-10
Add rancher istio v1.1.5 chart

guangbochen:
https://github.com/rancher/rancher/issues/19087


charts/rancher-istio/v1.1.1/charts/grafana/templates/deployment.yaml
```diff
@@ -22,7 +22,8 @@ spec:
         scheduler.alpha.kubernetes.io/critical-pod: ""
     spec:
       securityContext:
-        runAsUser: 472
+        runAsNonRoot: false
+        runAsUser: 0
```
gitlawr: Why does it need to be run as root, can we set runAsUser for each container instead of running as UID 0?

charts/rancher-istio/v1.1.1/charts/prometheus/templates/deployment.yaml
```diff
@@ -68,6 +68,31 @@ spec:
...
+          - containerPort: 80
+            name: http
+            protocol: TCP
+          {{- if and .Values.resources .Values.resources.proxy }}
+          resources:
+      {{ toYaml .Values.resources.proxy | indent 6 }}
+          {{- end }}
+          securityContext:
+            runAsNonRoot: false
+            runAsUser: 0
```
gitlawr: Ditto

charts/rancher-istio/latest/values.yaml
```diff
@@ -0,0 +1,485 @@
...
+certmanager:
+  enabled: false
+
+# Common settings used among istio subcharts.
+global:
+  # Default hub for Istio images.
+  # Releases are published to docker hub under 'istio' project.
+  # Daily builds from prow are on gcr.io, and nightly builds from circle on docker.io/istionightly
+  # hub: gcr.io/istio-release
+  hub: docker.io/istio
```
gitlawr, line 112: It should adapt to existing airgap framework for system-chart. Use `global.systemDefaultRegistry` in short. see other system-chart apps.

gitlawr:
We should move the chart to a specific version path and not use the `latest`

charts/rancher-istio/v1.1.1/charts/grafana/templates/deployment.yaml
orangedeng: I think it is used by nginx. The nginx needs to listen port 80 and this requires root privileges. I think we can use `setcap` to add privileges to nginx but it needs to customize the nginx image.

guangbochen:
moved to https://github.com/rancher/system-charts/pull/44

PR #43, opened 2019-05-13
Adjust Monitoring dashboards

thxCode:
**Problem:**
- The `Flunetd` pane from `Rancher Components` dashboard cannot show the right counting of fluentd Pods
- There are two kubelet scraping targets on Prometheus, one is scraping `/metrics`, another one is scraping `/metrics/cadvisor`. The metrics from `/metrics` endpoint are not including `container_name`. So the `container_*` expression will double the actual mount without
`conatiner_name!=""`

**Solution:**
- Change `sum(kube_pod_info{pod=~"fluentd.*"})` to `sum(kube_pod_info{pod=~".*fluentd.*",pod!~".*aggregator.*"})`
- Add `container_name!=""` into the expression

**Issue:**
- rancher/rancher#19722
- rancher/rancher#20162


loganhz:
It's not easy to review here. But the PR is adding the missing `conatiner_name!=""` back in some Grafana dashboard.

LGTM

PR #44, opened 2019-05-14
Add rancher istio v1.1.5 chart

orangedeng:
> TODO update to related images when the image is pushed into rancher repo

Related issue:
https://github.com/rancher/rancher/issues/19087

Related PR:
https://github.com/rancher/system-charts/pull/42


charts/rancher-istio/1.1.5/templates/sidecar-injector-configmap.yaml
```diff
@@ -0,0 +1,333 @@
...
+      rewriteAppHTTPProbe: {{ .Values.sidecarInjectorWebhook.rewriteAppHTTPProbe }}
+{{- if or (not .Values.istio_cni.enabled) .Values.global.proxy.enableCoreDump }}
+      initContainers:
+      {{ "[[ if ne (annotation .ObjectMeta `sidecar.istio.io/interceptionMode` .ProxyConfig.InterceptionMode) \"NONE\" ]]" }}
+{{- if not .Values.istio_cni.enabled }}
+      - name: istio-init
+{{- if contains "/" .Values.global.proxy_init.image }}
+        image: "{{ .Values.global.proxy_init.image }}"
+{{- else }}
+        image: "{{ .Values.global.hub }}/{{ .Values.global.proxy_init.image }}:{{ .Values.global.tag }}"
```
gitlawr: I can still see this `global.hub` being used. Also, check some other places

charts/rancher-istio/1.1.5/charts/certmanager/templates/deployment.yaml
```diff
@@ -36,7 +36,7 @@ spec:
 {{- end }}
       containers:
       - name: certmanager
-        image: "{{ .Values.hub }}/cert-manager-controller:{{ .Values.tag }}"
+        image: "{{ .Values.global.systemDefaultRegistry }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
gitlawr: We shouldn't do this. When `systemDefaultRegistry ` is empty, it is rendered as "/rancher/image"

charts/rancher-istio/1.1.5/values.yaml
```diff
@@ -250,7 +280,9 @@ global:
 
   proxy_init:
     # Base name for the proxy_init container, used to configure iptables.
-    image: proxy_init
+    # TODO update to rancher/istio-proxy_init when the image is pushed into rancher repo
```
gitlawr: Put the TODO to the PR description. Do we mirror chart images manually at the moment? I wonder if we can traverse system-chart values.yaml and extract images according to the keys in our mirror scripts.

test/ct.yaml
```diff
@@ -0,0 +1,8 @@
+remote: rancher-charts
```
gitlawr: Why is this file included?

orangedeng:
Updated

charts/rancher-istio/1.1.5/charts/prometheus/templates/deployment.yaml
```diff
@@ -68,6 +67,31 @@ spec:
             mountPath: /etc/prometheus
           - mountPath: /etc/istio-certs
             name: istio-certs
+        - name: prometheus-nginx
+          command:
+          - /bin/sh
+          - -c
+          - cp /nginx/run-sh.tmpl /var/run/nginx-start.sh; chmod +x /var/run/nginx-start.sh; /var/run/nginx-start.sh
```
gitlawr: From what I see, we are not using `run-sh.tmpl` as a template file, can we simply mount a `nginx-start.sh` file to the container and gives it executable permission. So that we don't need to do copy and chmod here, but simply use `/var/run/nginx-start.sh` as the command.

charts/rancher-istio/1.1.5/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: rancher-istio
-version: 1.1.1
-appVersion: 1.1.1
+version: 1.1.5
+appVersion: 1.1.5
```
gitlawr: Should we distinguish the chart version from appVersion? It would be trouble if we need to release a new chart without an appVersion change.

charts/rancher-istio/1.1.5/charts/prometheus/templates/deployment.yaml
orangedeng: I think we should keep this style until we tested the shell script mount from configmap can run successfully.

gitlawr:
LGTM

loganhz:
Link to https://github.com/rancher/rancher/issues/19087

loganhz:
Close in favor of https://github.com/rancher/system-charts/pull/51

PR #46, opened 2019-05-15
Add CI scripts

gitlawr:
Do Helm lint validation and Helm unittest for the charts

It uses the same test tools as the rancher chart, but a newer version of unittest plugin to address some issues of it.
It does linting for all charts, and run unit tests when a `tests` directory is found in a chart directory.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,13 @@
+kind: pipeline
+name: default
+
+steps:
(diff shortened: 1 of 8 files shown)


guangbochen:
LGTM.

cjellick:
LGTM

PR #47, opened 2019-05-17
Keep the operator charts version to 0.0.2

orangedeng:
**Problem:**
Upgrade the operator would have errors because we change the version of
the operator charts.

**Solution:**
Keep the old version 0.0.2 the operator used before.

Related issue:
https://github.com/rancher/rancher/issues/20301


guangbochen:
LGTM.

loganhz:
There are some changes in operator. So it should be a new version

PR #52, opened 2019-05-30
Update logging toleration to tolerate everything

aiwantaozi:
Problem:
Tolerate NoSchedule couldn't schedule logging to all node by default
Solution:
Update fluentd daemonset to tolerate everything by default

Issue:
https://github.com/rancher/rancher/issues/20565


alena1108:
@aiwantaozi we should make it conditional. And preserve the old behavior as a default way, and let user enable logging for etcd/control plane on demand, using some param on logging that gets translated to the catalog answer.

PR #54, opened 2019-06-03
Add Jaeger and grafana external url of Kiali

guangbochen:
**Problem:**
User is not able to view external Jaeger or Grafana URL in the Kiali dashboard, it shows a Rancher 404 page.

**Solution:**
as per official reply from https://github.com/istio/istio.io/issues/4146#issuecomment-493543032, the current solution is to add rancher generate Jaeger proxy url to the Kiali and Grafana config.

**Related Issue:**
https://github.com/rancher/rancher/issues/20615


loganhz:
Please do the same for Grafana 

https://52.221.207.19/k8s/clusters/c-btl7x/api/v1/namespaces/istio-system/services/http:grafana:80/proxy/

guangbochen:
updated.

loganhz:
LGTM

cjellick:
LGTM

PR #56, opened 2019-06-05
[WIP] add support to use cluster-monitoring of istio app

guangbochen:
**Note:**
This is a TEST-only PR, will need to split into two separated PR of `istio` and `monitoring` of submission.

---

**Issues/Enhancement:**
User should be able to re-use existing monitoring services(e.g, cluster-level monitoring Prometheus and Grafana) of their istio app.

**Solution:**
`Monitoring:`
1. Add default istio scrape rules to the cluster-monitoring Prometheus 
2. Add default istio dashboard to the cluster-monitoring Grafana. 

`Istio:`
1. auto config cluster-monitoring settings(i.e, Prometheus and Grafana endpoints) in istio Kalia when the monitoring type is set to `global.monitoring.type=cluster-monitoring`.

---

**TODO Notes:**
1. Rancher/Rancher istio traffic metrics should set its `Prometheus` endpoint to cluster-monitoring or external services when not using default monitoring service.


guangbochen:
**verification steps:**

1. change the default `system-library` url and branch at the `Global -> Tools -> Catalog` menu, e.g,  https://github.com/guangbochen/system-charts with `istio-moni` branch.
2. enable the cluster level monitoring with the default configuration.
3. enable the cluster level `service-mesh` with set the advanced configuration `global.monitoring.type=cluster-monitoring`
4. There should be no Grafana and Prometheus deployed in the istio deployment apps.
5. deploy the [istio bookinfo app](https://github.com/guangbochen/istio-chart), and check both istio Kiali dashboard and cluster-monitoring Grafana dashboard works as expected and contains the metrics data.

PR #58, opened 2019-06-06
Set maxUnavailable to 1 for registry deployment

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/20664

Problem:
Registry pod stuck in AttchVolumeFailed status on upgrade. Deadlock happens where the new pod waits the old pod to dettach the volume, and the old one waits for the new one to be ready.

Solution:
Set maxUnavailable to 1, so that the old pod can be released before the new pod is ready.


guangbochen:
LGTM.

PR #61, opened 2019-06-12
Add persistence.affix option to Harbor chart

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/20694

Problem:
Global registry uses existing PVCs from previous release, even storageclass changed on
re-enabling. Because the PVC names remain the same. In helm style, release name is usually used to avoid resources
conflicts. But we want to keep app name consistent and the release name
is bound to the app name.

Solution:
Add the option to specify affix for PVC names. UI can set a random
alphabet-number string to it on enabling, and keep it on updates. So
that every time registry is enabled, fresh PVCs are used.


guangbochen:
LGTM.

gitlawr:
Per https://github.com/rancher/rancher/issues/20694#issuecomment-501862669, we should delete PV/PVCs on disabling, so we don't need this identifier.

PR #65, opened 2019-06-17
fix versioning in istio

deniseschannon:
Per helm docs, version needs to be semver, appversion can be whatever we want. 

We should minimally match appversion with the folder version


cjellick:
I assume it wwas 1.1.5 because that is where we forked from upstream. Are we comfortable losing that connection?

cjellick:
I assume it wwas 1.1.5 because that is where we forked from upstream. Are we comfortable losing that connection?
I see that the app version remains 1.1.5, so maybe that is enough

cjellick:
LGTM

PR #71, opened 2019-07-08
Support windows logging

aiwantaozi:
Issue:
https://github.com/rancher/rancher/issues/20510


orangedeng:
I think you need to squash commits.

aiwantaozi:
updated, need to update windows image, after it's released

charts/rancher-logging/0.1.2/charts/fluentd-tester/templates/deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+        chart: {{ template "fluentd-tester.version" . }}
+        release: {{ .Release.Name }}
+        heritage: {{ .Release.Service }}
+    {{- if .Values.labels }}
+    {{ toYaml .Values.labels | indent 4 }}
+    {{- end }}
+{{- if .Values.annotations }}
+{{ toYaml .Values.annotations | indent 8 }}
+{{- end }}
+    spec:
```
orangedeng: Do you think fluentd-tester should have node selector for linux also?

charts/rancher-logging/0.1.2/charts/fluentd-tester/values.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+serviceAccount:
+  create: true
+
+annotations: {}
+
+# updateStrategy:
+#   type: RollingUpdate
+
+tolerations:
```
orangedeng: Where is the toleration key and value?

charts/rancher-logging/0.1.2/charts/fluentd/values.yaml
```diff
@@ -0,0 +1,395 @@
...
+service:
+  type: ClusterIP
+  ports:
+    - name: metric
+      port: 24231
+      targetPort: metric
+
+secrets:
+  entry:
+    fluent.conf: |-
```
orangedeng, line 24: Can we move all the secret data to other files instead of `values.yaml`

charts/rancher-logging/0.1.2/charts/fluentd-tester/values.yaml
aiwantaozi: Empty key with operator Exists matches all keys, values and effects which means this will tolerate everything.

charts/rancher-logging/0.1.2/charts/fluentd/values.yaml
aiwantaozi: Place these configures in value in order to allow user change them, if you place these configures in directly in secret, after redeploy the app, user's configure will be overwrite by default.

aiwantaozi:
updated

orangedeng:
LGTM

aiwantaozi:
updated node schedule to base on kubenetes version

orangedeng:
LGTM

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/templates/daemonset.yaml
```diff
@@ -0,0 +1,169 @@
...
+          name: ssl
+        - mountPath: {{ .Values.cluster.dockerRoot }}
+          name: dockerroot
+        - mountPath: /var/log/containers
+          name: varlogcontainers
+        - mountPath: /var/log/pods
+          name: varlogpods
+        # - mountPath: /var/lib/rancher/rke/log
+        #   name: rkelog
+        # - mountPath: /var/lib/rancher/log-volumes
```
gitlawr: Why do we comment out these volume mounts, shall we delete them? Also, check other comment blocks.

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/templates/daemonset.yaml
```diff
@@ -0,0 +1,169 @@
...
+{{- end }}
+{{- if .Values.livenessProbe.enabled }}
+        # Liveness probe is aimed to help in situarions where fluentd
+        # silently hangs for no apparent reasons until manual restart.
+        # The idea of this probe is that if fluentd is not queueing or
+        # flushing chunks for 5 minutes, something is not right. If
+        # you want to change the fluentd configuration, reducing amount of
+        # logs fluentd collects, consider changing the threshold or turning
+        # liveness probe off completely.
+        # TODO
```
gitlawr: Do we plan to address this?

charts/rancher-logging/0.1.2/values.yaml
```diff
@@ -4,6 +4,5 @@ fluentd-tester:
   enabled: false
 log-aggregator:
   enabled: false
```
gitlawr, line 6: Conditions in requirements.yml has changed to `log-aggregator.log-aggregator-<OS>.enabled`, I think we should update them in the values as well

aiwantaozi: this values.yaml file is for the first level sub charts, first level sub charts only have fluentd, fluentd-test, log-aggregator. <OS> is included in the second level sub charts.

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/templates/daemonset.yaml
aiwantaozi: addressed

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/templates/daemonset.yaml
aiwantaozi: these will support later

aiwantaozi:
@gitlawr Just updated, thanks

gitlawr:
LGTM

aiwantaozi:
rebased

aiwantaozi:
update windows livenessProbe and remove label chart in matchLabels. Because label chart include version, which will be changed after we release a new version, and in v1 if we are going to update matchLabels will get an error. ```Forbidden: updates to deployment/statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden]```

loganhz:
LGTM

aiwantaozi:
rebase to resolve conflicts

loganhz:
Link to https://github.com/rancher/rancher/issues/20123

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/values.yaml
```diff
@@ -0,0 +1,50 @@
...
+  ports:
+    - name: metric
+      port: 24231
+      targetPort: metric
+
+configmapReload:
+  name: reloader
+  image:
+    repository: micheliac/configmap-reload
+    tag: dev-windows-1809
```
loganhz: Please update

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/values.yaml
```diff
@@ -0,0 +1,50 @@
+nameOverride: fluentd
+labels: {}
+
+image:
+  repository: micheliac/fluentd
+  tag: dev-windows-1809
```
loganhz: Please update

charts/rancher-logging/0.1.2/charts/log-aggregator/charts/log-aggregator-windows/values.yaml
```diff
@@ -0,0 +1,27 @@
+nameOverride: log-aggregator
+
+labels: {}
+## Log Aggregator container image
+##
+image:
+  repository: micheliac/log-aggregator
+  tag: windows
```
loganhz: Please update

loganhz:
lgtm

cjellick:
@cloudnautique or @drpebcak can you review?

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-linux/templates/daemonset.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: {{ template "daemonset_api_version" . }}
 kind: DaemonSet
 metadata:
-  name: {{ template "fluentd.fullname" . }}
+  name: {{ template "fluentd.fullname" . }}-linux
```
cjellick, line 4: What is the upgrade impact of change the name of this daemonSet?

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/templates/daemonset.yaml
cjellick: I'm really not a fan of committing commented out. A dev looking at this a few months from now won't know why this is included but commented out. Please remove.

cjellick: same applies to other places in the code

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-linux/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+name: fluentd-linux
+version: 0.0.1
+appVersion: 1.1.3
```
drpebcak: These versions don't line up with the versions mentioned in the directory structure

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+name: fluentd-windows
+version: 0.0.1
+appVersion: 1.1.3
```
drpebcak: These versions don't line up with the versions mentioned in the directory structure

charts/rancher-logging/0.1.2/charts/log-aggregator/charts/log-aggregator-linux/templates/daemonset.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: {{ template "daemonset_api_version" . }}
 kind: DaemonSet
 metadata:
-  name: {{ template "log-aggregator.fullname" . }}
+  name: {{ template "log-aggregator.fullname" . }}-linux
```
drpebcak, line 4: I don't think you can change the name of a daemonset on upgrade.

charts/rancher-logging/0.1.2/charts/log-aggregator/charts/log-aggregator-windows/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+name: log-aggregator-windows
+version: 0.0.1
```
drpebcak, line 2: These versions don't line up with the versions mentioned in the directory structure

drpebcak (changes requested):
I'm not sure I have the full context here as to why these need to be separate charts, but in general I think the version numbers are a bit confusing. Is there a reason none of the chart versions line up?

charts/rancher-logging/0.1.2/charts/log-aggregator/charts/log-aggregator-linux/templates/daemonset.yaml
aiwantaozi: Because we separate to Linux and Windows, so we add a suffix for that Daemonset, we use force upgrade which will remove the old chart and install the new chart. Because some changes between two version charts may cause upgrade failed, so we use force upgrade, for example, we remove label chart under `spec.selector.matchLabels`, and matchLabels is an unmodifiable field in `apps/v1` Daemonset.

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/templates/daemonset.yaml
aiwantaozi: removed

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-linux/templates/daemonset.yaml
aiwantaozi: @cjellick  Because we separate to Linux and Windows, so we add a suffix for that Daemonset, we use force upgrade which will remove the old chart and install the new chart. Because some changes between two version charts may cause upgrade failed, so we use force upgrade, for example, we remove label chart under spec.selector.matchLabels, and matchLabels is an unmodifiable field in apps/v1 Daemonset.

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-linux/Chart.yaml
aiwantaozi: Hi @drpebcak we have a version for rancher-logging charts, which is under `rancher/system-charts/charts/rancher-logging/0.1.2/Chart.yaml`, for sub-charts fluentd/fluentd-tester/log-aggregator, if the charts under their directory changed, then we change the version. For sub-charts like `rancher-logging/0.1.2/charts/fluentd/charts/fluentd-linux` if itself changed, then we change version. The appVersion is bind to the fluentd version, so I change it to `1.6.3` and update some charts version to 0.0.2. Is this ok?

charts/rancher-logging/0.1.2/charts/log-aggregator/charts/log-aggregator-linux/templates/daemonset.yaml
cjellick: Have you explicitly tested upgrade scenarios around this and verified it works as expected for linux? I need confirmation on that before merging.

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-linux/Chart.yaml
drpebcak: Makes sense. I see now that 0.1.2 is not a released version of the `rancher-logging` chart

cjellick:
@loganhz @aiwantaozi need to answer this question from @drpebcak 
> I'm not sure I have the full context here as to why these need to be separate charts, but in general I think the version numbers are a bit confusing. Is there a reason none of the chart versions line up?

Specifically, why do you need two separate charts for linux vs windows? Can you have a single chart controlled by templating? If there's a good reason, just let us know.

charts/rancher-logging/0.1.2/charts/log-aggregator/charts/log-aggregator-linux/templates/daemonset.yaml
aiwantaozi: Yes, I have tested the upgrade scenarios.

aiwantaozi:
@cjellick Because of values.yaml file for Windows and Linux is different if we use one template for both of them, rancher-logging will be separate info two apps and the deployment and upgrade logic is a bit different.

cjellick (approved):

drpebcak (approved):

PR #74, opened 2019-07-23
Do not restrict the files name to letters and numbers in fluentd

aiwantaozi:
Problem:
now filename only allow numbers and letter

Solution:
change regex expression to allow all charaters

Issue:
https://github.com/rancher/rancher/issues/17022


charts/rancher-logging/0.1.2/charts/fluentd/values.yaml
```diff
@@ -143,22 +143,22 @@ secrets:
...
-          format ${component = tag.match(Regexp.compile('tmp-cluster-custom\.var\.lib\.rancher\.log-volumes\.(?<pod_id>[^_]+)_(?<volume_name>[^_]+)\.(?<format>[a-z0-9]+)\.(?<cluster_id>[^_]+)_(?<cluster_name>[^_]+)_(?<namespace>[^_]+)_(?<project_id>[^_]+)_(?<project_name>[^_]+)_(?<workload_name>[^_]+)_(?<pod_name>[a-z0-9]([-a-z0-9]*[a-z0-9])+)_(?<container_name>[^_^.]+)\.(?<filename>[a-z0-9A-Z]+\.[a-zA-Z0-9]+)')); component['format']}
+          cluster_id ${component = tag.match(Regexp.compile('tmp-cluster-custom\.var\.lib\.rancher\.log-volumes\.(?<pod_id>[^_]+)_(?<volume_name>[^_]+)\.(?<format>[a-z0-9]+)\.(?<cluster_id>[^_]+)_(?<cluster_name>[^_]+)_(?<namespace>[^_]+)_(?<project_id>[^_]+)_(?<project_name>[^_]+)_(?<workload_name>[^_]+)_(?<pod_name>[a-z0-9]([-a-z0-9]*[a-z0-9])+)_(?<container_name>[^_^.]+)\.(?<filename>[^^]+)')); component['cluster_id']}
```
gitlawr: Are we excluding the `^` character here(`[^^]+`)? I think it is valid in a filename.

aiwantaozi: Yes, because caret is invalid in some filesystem, like Windows using the FAT file system

gitlawr: I see

gitlawr:
LGTM

charts/rancher-logging/0.1.2/charts/fluentd/values.yaml
gitlawr: If it is invalid, we don't need to care about it so as to match the filenames. Doing that will break when the character is valid.

gitlawr:
LGTM

cjellick:
lgtm

PR #75, opened 2019-07-23
Fixed issues encountered when deploying using restricted psp

MrBones757:
When deploying to an environment with a restricted psp (CIS hardened as per documentation), the monitoring cluster fails to start.

During the debugging of this issue the following changes were identified as required in our environment. These changes included:
- Adding support for custom PSP for exporter-node with applicable ClusterRoles and ClusterRoleBindings
- Fixed type in api verb (creat -> create)
- Swicthed NGINX to use 8080 instead of 80 to allow non-root execution
- Moved the nginx pid file to a directory writable by NGINX user.
- Applied UID/GID/FSGROUP for various services.

If someone could review and second-verify these changes that would be great. Unfortunately i cant test these in a default cluster / any other environment as i'm limited by the available infrastructure.


MrBones757:
There appears to be an issue with this following a node reboot where nginx fails to chown /tmp/nginx. Possibility to change the nginx directory to somethign it already owns would work. Ill trial this fix in my environment

MrBones757:
The changes in the lates commit fix the prometheus-proxy start issues.
- moved to port 80
- changed uid to 100/101 on proxy
- updated nginx config to use paths owned by the nginx user rather than root to avoid perm issues when not running as uid 0 / root. Deployed successfully in additional testing environment/s.

Apologies for the weird file re-order changes... it appears a combination of using bash for windows & vscode plugins decided to redo both that and the permissions.

loganhz:
Hi @MrBones757 could you rebase the change into monitoring v0.0.4, please?

wallies:
Can validate I had this issue on Kubernetes 1.15, Rancher 2.2.9, using this version of the chart. This fixed the issue. I am also using restricted psp

recena:
Closing due lack of activity. Old content.

PR #82, opened 2019-08-26
Add host path variable

xiaoluhong:
In a production environment, most users use external mounted disks to store data, so the host path is not necessarily the standard path. This PR specifies the host path through a variable, allowing the user flexibility to modify it.


xiaoluhong:
@aiwantaozi

PR #83, opened 2019-08-27
Improve linux node selector and tolerations

aiwantaozi:
1. Improve linux node selector and tolerations
2. upgrade wmi exporter image

### DIFF
--- a/charts/rancher-monitoring/v0.0.4/templates/_helpers.tpl
+++ b/charts/rancher-monitoring/v0.0.4/templates/_helpers.tpl
@@ -137,3 +137,11 @@ add below linux tolerations to workloads could be scheduled to those linux nodes
   effect: "NoSchedule"
   operator: "Equal"
 {{- end -}}
+
+{{- define "linux-node-selector" -}}
+{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+beta.kubernetes.io/os: linux
+{{- else -}}
+kubernetes.io/os: linux
+{{- end -}}
+{{- end -}}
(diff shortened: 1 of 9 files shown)


loganhz:
lgtm

loganhz:
link to 
https://github.com/rancher/rancher/issues/20122
https://github.com/rancher/rancher/issues/20077

charts/rancher-monitoring/v0.0.4/templates/_helpers.tpl
```diff
@@ -136,4 +136,12 @@ add below linux tolerations to workloads could be scheduled to those linux nodes
   value: "linux"
   effect: "NoSchedule"
   operator: "Equal"
-{{- end -}}
\ No newline at end of file
+{{- end -}}
+
+{{- define "linux-node-selector" -}}
+{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.GitVersion -}}
```
deniseschannon, line 142: we are only supporting k8s 1.15 so why does this logic exist/need to change?

aiwantaozi: Because for windows cluster, we need to schedule all the monitoring components to Linux nodes, so we need `nodeSelector`. After k8s 1.14, label `beta.kubernetes.io/os` is deprecated, use the label `kubernetes.io/os` instead. When a user has a k8s 1.13 cluster with monitoring enabled, and upgrade server to latest, then user could upgrade monitoring, if we only have `kubernetes.io/os` label, it will schedule fail. So for old version cluster, we can use the label `beta.kubernetes.io/os` or empty nodeSelector, I prefer the first way, in this way only kubeVersion affect nodeSelector. Even we not only support k8s 1.15, the monitoring still work well, still could be scheduled to Linux nodes.

deniseschannon: @cjellick I forgot that we were allowing k8s 1.14 to be in the product for now (for upgrade testing from k8s 1.13 to 1.14.

cjellick (approved):
LGTM

PR #84, opened 2019-08-28
Adapt monitoring to restricted PSP clusters

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/20229

Problem:
Fail to enable monitoring on restricted PSP clusters.

Solution:
1. Update grafana to run as grafana user.
2. Update nginx sidecars to run as nginx user, move files to the user
home directory(/var/cache/nginx) and listen on non-privileged 8080 port.
With 1 & 2, project monitoring works in restricted clusters out of the
box.
3. Add necessary PSP/role/rb resource to node-exporter, which is used by
cluster monitoring.
The added PSP is bound to node-exporter SA. Its rules include: restricted rules + hostPath + hostNetwork + hostPorts + hostPID

### DIFF
--- a/charts/rancher-monitoring/v0.0.4/charts/exporter-node/values.yaml
+++ b/charts/rancher-monitoring/v0.0.4/charts/exporter-node/values.yaml
@@ -1,4 +1,5 @@
 enabledRBAC: true
+enabledPSP: true
 
 ##
 # Default collector settings
--- a/charts/rancher-monitoring/v0.0.4/charts/exporter-node/templates/psp.yaml
+++ b/charts/rancher-monitoring/v0.0.4/charts/exporter-node/templates/psp.yaml
@@ -1,0 +1,44 @@
+{{- if .Values.enabledPSP }}
+apiVersion: extensions/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: {{ template "app.fullname" . }}
+spec:
+  allowPrivilegeEscalation: false
+  fsGroup:
+    ranges:
+    - max: 65535
+      min: 1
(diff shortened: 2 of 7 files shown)


charts/rancher-monitoring/v0.0.4/charts/exporter-node/values.yaml
```diff
@@ -1,4 +1,5 @@
 enabledRBAC: true
+enabledPSP: true
```
orangedeng, line 2: Do we need to change Rancher code also? is it ok to always enable PSP?

gitlawr: PSP is available from v1.5 and the API remains to be beta from 1.11 to 1.15(haven't check earlier ones).
It doesn't matter if the admission controller is enabled or not, these resources can be managed in the app anyway.
So it is ok to always turn this flag on.

orangedeng:
LGTM

cjellick:
@cloudnautique or @drpebcak can you review

cjellick:
I need to talk through this with @cloudnautique tomorrow. This is the opposite of what I suggested. Instead of leveraging the PSP features of rancher, we just put the raw psp resources into this helm chart. That means they are completely outside the purview of Rancher PSP management, which effectively defeats the purpose of the PSP feature in rancher.

cjellick:
Discussed the implementation with @cloudnautique and we are good with it. I have opened this enhancement so that Rancher will do a better job of surfacing PSP bindings in general https://github.com/rancher/rancher/issues/22525

PR #88, opened 2019-09-09
change fluentd windows init container command

aiwantaozi:
Problem:
windows will throws error when the file already exist, which is different from linux

Solution:
change to use new-item

Issue:
https://github.com/rancher/rancher/issues/22679


thxCode:
LGTM

thxCode (approved):

cbron (approved):
Seems fine to me. [Docs](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/new-item)

PR #89, opened 2019-09-09
Adapt logging to psp restricted clusters

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/22593

Add PSP and related role/rolebinding/SA resources to logging chart.
The PSP contains restricted rules + hostPath volume.
Also, remove unnecessary privileged escalation from log-aggregator.


aiwantaozi:
Could you check when `unrestricted-psp` created. Is unrestricted-psp always exist, even a cluster not enable psp?

gitlawr:
@aiwantaozi Updated to include the PSP resource in the chart. It is a common practice in community charts: https://github.com/helm/charts/search?q=PodSecurityPolicy&unscoped_q=PodSecurityPolicy
It is enabled by default so that it works no matter PSP is activated or not.

gitlawr:
The restricted PSP does not allow flexVolume usage, which is needed when users configure custom log collection volumes for their workloads. So users need to grant additional PSP rules on demand if they want to use that feature. i.e.
```
# Other PSP spec
  volumes:
    - flexVolume
  allowedFlexVolumes:
    - driver: cattle.io/log-aggregator
```
Would like to have @cloudnautique 's input to see if it makes sense. Or we should add the above rules to the restricted rules.

aiwantaozi:
@gitlawr Could you also update chart `fluentd-tester`

gitlawr:
@aiwantaozi The `fuentd-tester` workload obeys the restricted PSP so no update is needed for it.

aiwantaozi:
LGTM

aiwantaozi (approved):

charts/rancher-logging/0.1.2/templates/psp.yaml
```diff
@@ -0,0 +1,39 @@
...
+  volumes:
+  - configMap
+  - emptyDir
+  - projected
+  - secret
+  - downwardAPI
+  - persistentVolumeClaim
+  - hostPath
+  allowedHostPaths:
+    - pathPrefix: /
```
cjellick, line 38: upon further consideration and consultation with @ibuildthecloud this is perhaps too broad. can this be narrowed down at all?

gitlawr: The following are the hostPaths mounted by fluentd/log-aggregator. flexVolumeDir & dockerRoot are configurable by helm values.
```
#flexVolumeDir by logaggregator
/usr/libexec/kubernetes/kubelet-plugins/volume/exec
#by fluentd
/var/lib/rancher/fluentd/etc/config/custom
#dockerRoot
/var/lib/docker
/var/log/containers
/var/log/pods
/var/lib/rancher/rke/log
/var/lib/rancher/log-volumes
/var/lib/rancher/fluentd/log
/usr/lib64
```

gitlawr: Probably narrow it down to 
/var/lib/rancher
/var/log/containers
/var/log/pods
/usr/lib64
${flexVolumeDir:-"/usr/libexec/kubernetes/kubelet-plugins/volume/exec"}
${dockerRoot:-"/var/lib/docker"}

PR #91, opened 2019-09-11
Sync dev and v2.3-dev branch

loganhz:
Sync dev and v2.3-dev branch


deniseschannon:
We sync this whenever we cut a RC. there is no reason to create a PR for it.

PR #92, opened 2019-09-11
Set max version for monitoring and istio charts

loganhz:
We would like to use Rancher max version and min version to control if we show the version in enable monitoring and istio page

https://github.com/rancher/rancher/issues/22788


deniseschannon:
Please hold off on merging until the UI is tested

deniseschannon:
LGTM

PR #96, opened 2019-09-12
Distinguish Windows images

thxCode:
When we scan the system charts, there is no way to find out if there are some Windows images for it. So we would like to add a new field to indicate the image is used in Windows worker.

**Problem:**
Could not distinguish Windows images from `values.yaml`

**Solution:**
- Add a new field `os` below field `image`
- Must indicate `os: windows` if the image is using in Windows worker

**Issue:**
https://github.com/rancher/rancher/issues/22463

### DIFF
--- a/charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/values.yaml
+++ b/charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/values.yaml
@@ -2,6 +2,7 @@ nameOverride: fluentd
 labels: {}
 
 image:
+  os: windows
   repository: rancher/fluentd
   tag: v0.1.16
   pullPolicy: IfNotPresent
@@ -44,6 +45,7 @@ service:
 configmapReload:
   name: reloader
   image:
+    os: windows
     repository: rancher/configmap-reload
     tag: v0.3.0-rancher2
     pullPolicy: IfNotPresent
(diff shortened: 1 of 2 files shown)


loganhz:
lgtm

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/values.yaml
```diff
@@ -44,6 +45,7 @@ service:
 configmapReload:
   name: reloader
   image:
+    os: windows
     repository: rancher/configmap-reload
```
mrajashree, line 49: @thxCode don't we need the same for the windows [log-aggregator](https://github.com/rancher/system-charts/blob/dev/charts/rancher-logging/0.1.2/charts/log-aggregator/charts/log-aggregator-windows/values.yaml#L6)

drpebcak (approved):
lgtm

charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/values.yaml
thxCode: `rancher/log-aggregator` will run on the Linux side. I think we could only care about those images running on the Windows side: https://github.com/rancher/system-charts/blob/dev/charts/rancher-logging/0.1.2/charts/fluentd/charts/fluentd-windows/values.yaml

PR #98, opened 2019-09-17
Upgrade rancher fluentd image version to v0.1.17

aiwantaozi:
Problem:
Fluentd will start creating hundreds of thousands of 0 byte buffer log files when the partition it is on runs out of space.
https://github.com/fluent/fluentd/issues/2593

Solution:
Upgrade gems version and release rancher/fluentd:v0.1.17

Issue:
https://github.com/rancher/rancher/issues/22689


loganhz:
lgtm

PR #99, opened 2019-09-18
Fix project level monitoring can not upgrade

guangbochen:
**Problem:**
Project level Prometheus can't be upgraded due to `spec.podMetadata.creationTimestamp in the body must be of type string: "null"`

**Solution:**
Remove `{{- if .Release.IsInstall }}` check of `creationTimestamp` since this is required for both update and create.

**Related Issue:**
https://github.com/rancher/rancher/issues/22915


loganhz:
lgtm

PR #111, opened 2019-10-15
Respect namespace label from the nginx ingres controller

lalinsky:
To allow project monitoring to access nginx ingress metrics, the correct value needs to be in the namespace field. Currently, it stored in exported_namespace, because the namespace label is already defined.


recena:
Closing due lack of activity. Old content.

PR #113, opened 2019-10-22
Fix nginx user permissions in nginx-alpine image

gitlawr:
Address issues: 
https://github.com/rancher/rancher/issues/23573

Problem:
Fails to run nginx proxies, because user home directory is not owned by nginx user inside newer nginx-alpine images

Solution:
Mount emptyDir to the user home directory(/var/cache/nginx) so that it
is writable by the nginx user.
Also update UID from 100 to 101 to be consistent with upstream changes(https://github.com/nginxinc/docker-nginx/issues/218)


thxCode:
LGTM

PR #102, opened 2019-09-21
Enable securityContext for exporter-kube-state and operator-init

SeanThomasWilliams:
Update securityContext for monitoring containers. Allows monitoring to run when Rancher hardening rules are applied.

https://github.com/rancher/rancher/issues/20884

PR #114, opened 2019-10-22
Update securityContext for monitoring containers

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/20884

Moved from https://github.com/rancher/system-charts/pull/102

To make monitoring work when hardening rules are applied.


loganhz:
lgtm

PR #117, opened 2019-10-31
[WIP] Configure healthcheck probes on monitoring services

thxCode:
**Problem:**
- The Prometheus initialization time cost should relate to the cluster size, sample retention period, storage I/O and so on, we could not suit different environments with a fixed value.
- The users could not configure the `kube-state-metrics` & `grafana` probes too.

**Solution:**
Make the `livenessProbe` and `readinessProbe` configurable. With this PR, the users could adjust the `initialDelaySeconds` of those two probes in `Advanced` options from the Monitoring page.


loganhz:
Close in favor of #138

PR #120, opened 2019-11-07
Compatible v1.16

thxCode:
**Problem:**
Losing some graphs in Rancher UI, when deploying in v1.16+

**Solution:**
Bump into v0.0.6:
- Compatible with v1.16 deprecated metrics
- Compatible with v1.16 deprecated labels

**Issue:**
https://github.com/rancher/rancher/issues/23731


charts/rancher-monitoring/v0.0.6/README.md
```diff
@@ -0,0 +1,8 @@
+# rancher-monitoring
+
+Installs [prometheus-operator](https://github.com/coreos/prometheus-operator) to create/configure/manage Prometheus clusters atop Kubernetes.
+
+> **Tip**: Only use for Rancher Monitoring!!!
+
+## Prerequisites
+  - Rancher 2.2+
```
orangedeng: nit. Should we declare that this chart is for k8s 1.16?

thxCode: No, we don't need to do that. This version could work in `<= v1.13` as well, we use a version comparison method (`semverCompare`) to compatible with the higher Kubernetes.

orangedeng (approved):
LGTM

PR #121, opened 2019-11-11
Add creationTimestamp to prometheus storage metadata

guangbochen:
**Problem:**
`spec.storage.volumeClaimTemplate.metadata.creationTimestamp` in body must be of type string: "null"' error upgrading cluster-monitoring 0.0.3 with persistent prometheus storage to cluster-monitoring 0.0.5

**Solution:**
Add creationTimestamp to prometheus storage metadata

**Related Issue:**
https://github.com/rancher/rancher/issues/23857


loganhz:
Close it as it can not fix the issue

PR #122, opened 2019-11-11
updates to CIS chart

leodotcloud:
Addresses:
https://github.com/rancher/rancher/issues/23904
https://github.com/rancher/rancher/issues/23905


deniseschannon:
Since we don't expect people to have been using v0.0.1, any reason why we can't just update v0.0.1?

PR #137, opened 2019-12-18
Wip: Update istio to 1.4.2

StrongMonkey:


StrongMonkey:
General questions: 
1. Do we need to update and add all istio versions? 
2. What's automating image mirroring from docker.io/istio to docker.io/rancher? Looks like we are using mirrored images under rancher org.

deniseschannon:
1. No, you don't need to add all versions. 
2. You add the updated images/tags to this file and it automatically pulls from istio and pushes it to docker.io/rancher: https://github.com/rancher/rancher/blob/c13661520987f911fb6020347339969860912805/scripts/istio-images-mirror

charts/rancher-istio/0.1.2/Chart.yaml
```diff
@@ -1,20 +1,17 @@
 apiVersion: v1
 name: rancher-istio
-version: 0.1.1
```
deniseschannon: We should match the helm chart version with the Istio version. It'll make it easier for users to understand.

charts/rancher-istio/0.1.2/Chart.yaml
```diff
@@ -1,20 +1,17 @@
...
-appVersion: 1.3.3
-tillerVersion: ">=2.7.2-0"
+version: 0.1.2
+appVersion: 1.4.2
 description: Helm chart for all istio components
-home: https://istio.io/
```
deniseschannon: what does eliminating this line do?

charts/rancher-istio/0.1.2/charts/certmanager/templates/deployment.yaml
```diff
@@ -59,7 +55,7 @@ spec:
       {{- end }}
       {{- if .Values.podDnsConfig }}
       dnsConfig:
-      {{ toYaml .Values.podDnsConfig | indent 8 }}
+{{ toYaml .Values.podDnsConfig | indent 8 }}
```
deniseschannon, line 58: Spacing is off.

charts/rancher-istio/0.1.2/charts/certmanager/values.yaml
```diff
@@ -5,20 +5,21 @@
 # restart, DestinationRules can be created using the ACME-signed certificates.
 enabled: false
 replicaCount: 1
-image:
-  repository: rancher/jetstack-cert-manager-controller
-  tag: v0.6.2
+hub: quay.io/jetstack
+image: cert-manager-controller
+tag: v0.8.1
```
deniseschannon, line 10: Should this be a later version of cert-manager?

charts/rancher-istio/0.1.2/charts/certmanager/values.yaml
```diff
@@ -5,20 +5,21 @@
 # restart, DestinationRules can be created using the ACME-signed certificates.
 enabled: false
 replicaCount: 1
-image:
-  repository: rancher/jetstack-cert-manager-controller
-  tag: v0.6.2
+hub: quay.io/jetstack
```
deniseschannon, line 8: Since we need all images to reside in dockerhub, what does changing this to quay.io do? I don't believe we should be updating it.

deniseschannon (changes requested):
I haven't reviewed the whole PR, but there are also some weird spacing for any of the templating being used/

StrongMonkey:
@deniseschannon All changes under charts/ are pulled directly from vanilla charts. I am trying to make this chart as much as possible to align with what istio upstream offers. The only changes I made are chart.yaml, question.yaml and values.yaml under the root directly of this chart. I will separate these changes into different commits so you will see what's actually changes and what's get copied from upsteam.

charts/rancher-istio/0.1.2/charts/certmanager/values.yaml
StrongMonkey: yeah, images need to be updated and that will be under values.yaml in root directory.

charts/rancher-istio/0.1.2/Chart.yaml
StrongMonkey: I agree. Previous istio version are used as 0.1.1 so I just make it more compatible.

deniseschannon: @StrongMonkey Maybe we shouldn't change it as the UI already displays it with the app version and the Istio version. 
Screen Shot 2019-12-20 at 9 46 42 AM

PR #138, opened 2019-12-20
Support read/write from/to remote endpoints and health check configuration

shanewxy:
This pull request involves two enhancements:

1. Enable configurations for Prometheus remote write and read endpoints(e.g. to Influxdb or VictoriaMetrics)  for cluster and project level monitoring.

- Related issue: https://github.com/rancher/rancher/issues/20624
- 截屏2019-12-20上午11 45 51

2. Allow for the configuration of the health check probes on monitoring services including Prometheus, grafana, and exporter-kube-state.

- Related issue: https://github.com/rancher/rancher/issues/23874


thxCode (dismissed):
LGTM

adeleglise:
This would be really great to have this !

shanewxy:
Updated to 0.0.8

loganhz (approved):

charts/rancher-monitoring/v0.0.8/values.yaml
```diff
@@ -376,6 +380,8 @@ prometheus:
     alertManagerNamespace: ""
   thanos:
     enabled: false
+  livenessProbe: {}
```
cloudnautique, line 383: Do the empty maps pass empty defaults to the corresponding charts?

charts/rancher-monitoring/v0.0.8/charts/prometheus/templates/prometheus.yaml
```diff
@@ -151,10 +151,13 @@ spec:
   serviceMonitorSelector:
 {{ toYaml .Values.serviceMonitorSelector | indent 4 }}
 {{- end }}
-{{- if and (eq .Values.level "project") (eq .Values.sync.mode "remote") }}
+{{- if .Values.remoteRead }}
   remoteRead:
-  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
-    url: {{ printf "%s%s" .Values.sync.target .Values.sync.path }}
+{{ toYaml .Values.remoteRead | indent 4 }}
+{{- end }}
```
cloudnautique, line 143: For the read  / write specs its a list of map. So the UX would be adding:

```
prometheus.remoteRead[0].url = http://remote
prometheus.remoteRead[0].bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```
and similarly for remoteWrite?
```

File: charts/rancher-monitoring/v0.0.8/values.yaml (hunk 1 above)
Comment (loganhz) [reply charts/rancher-monitoring/v0.0.8/values.yaml:383]: We are using `{{ merge .Values.livenessProbe .Values.defaultLivenessProbe | toYaml | indent 6 }}` in chart. We will merge `defaultLivenessProbe` and the empty (if users don't config it) `livenessProbe`

File: charts/rancher-monitoring/v0.0.8/charts/prometheus/templates/prometheus.yaml (hunk 2 above)
Comment (loganhz) [reply charts/rancher-monitoring/v0.0.8/charts/prometheus/templates/prometheus.yaml:143]: Hi @cloudnautique 

The question is what fields we would like to expose for remote read and remote write in Rancher UI. 

We only exposed remote write url in the UI PR. Other fields such as `bearerTokenFile` need to be configured in custom answers section. 

We can do the same for remote read. Is that OK if we only expose URL field (URL field is the only required field for remote read and write)?

https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#remotewritespec
https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#remotereadspec

Comment (cloudnautique) [reply charts/rancher-monitoring/v0.0.8/charts/prometheus/templates/prometheus.yaml:143]: My question is what would the UX look like. My fear is the user will need to keep tabs on the index... 

In the UI if they use:
```
url1: https://x.y
url2: https://z.y
```

then they would need to know that
`prometheus.remoteRead[0]` applies to the first url.

It is kind of ugly. If we needed to do that then, I would want to see the index number in the UI.

charts/rancher-monitoring/v0.0.8/values.yaml
cloudnautique: I see, so the merge function just allows the user to override a specific field without having to pass the entire map?

charts/rancher-monitoring/v0.0.8/charts/prometheus/templates/prometheus.yaml
cloudnautique: @loganhz, I do not want the UI if it only shows the URL box. Our experience has been that TLS configs and basic auth are going to be more common when it comes to these things. Having to use two separate input methods feels wrong.

I would accept a UI field that took a complete remote{read|write}spec or just use the advanced options.

That said, I would like to see what a user would enter for a remotereadspec with a URL and TLS/Auth configuration.

charts/rancher-monitoring/v0.0.8/values.yaml
loganhz: Yep

charts/rancher-monitoring/v0.0.8/charts/prometheus/templates/prometheus.yaml
loganhz: Make sense! We will close the UI PR. So users could just use the advanced options for now.

fsperling: Hey,

we use multiple remote_write urls, basic auth and also the queue_config fields.
I'd be fine with have one field for the whole remote write spec.

dramich:
The changes look fine but curious if health check probes is going to be an API only thing? Also there should be a docs issue or update related to the changes.

dramich (approved):

loganhz:
Docs issue https://github.com/rancher/docs/issues/2276

kamushadenes:
Despite https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#basicauth, I can't get basic auth working.
It either doesn't do anything when I save, or simply doesn't deploy prometheus-cluster-monitoring if I insert the data before enabling monitoring.

prometheus.remoteWrite[0].basicAuth.password: SOME_BASE64=
prometheus.remoteWrite[0].basicAuth.username: USERNAME
prometheus.remoteWrite[0].name: production
prometheus.remoteWrite[0].url: https://url

If I just use the URL, it works, but obviously fails to authenticate.

Is this a known issue?

PR #143, opened 2020-01-11
Fix kiali graph, add istio 1.4.3

StrongMonkey:


cbron:
Switching to: https://github.com/rancher/system-charts/pull/145

PR #147, opened 2020-01-13
Update minimum rancher version in Istio charts

daxmc99:
Due to changes in the UI of the Istio chart we need to add minimum
Rancher versions to ensure that the Istio chart will display its UI
elements correctly within Rancher.

### DIFF
--- a/scripts/update-istio-release
+++ b/scripts/update-istio-release
@@ -45,11 +45,14 @@ EOF
 done
 rm -r charts/rancher-istio/${1}/charts/istio-init
 
-# Add question.yaml
+# Add question.yaml, make need to update minimum rancher version
+RANCHER_MIN_VERSION="2.3.4-rc1"
+echo "Current rancher min version is set to $RANCHER_MIN_VERSION"
+
 cat > charts/rancher-istio/${1}/questions.yaml << EOF
 labels:
   rancher.istio.v${1}: ${1}
-rancher_min_version: 2.3.0-rc1
+rancher_min_version: $RANCHER_MIN_VERSION
 EOF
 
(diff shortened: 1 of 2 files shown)


scripts/update-istio-release
```diff
@@ -49,7 +49,7 @@ rm -r charts/rancher-istio/${1}/charts/istio-init
 cat > charts/rancher-istio/${1}/questions.yaml << EOF
 labels:
   rancher.istio.v${1}: ${1}
-rancher_min_version: 2.3.0-rc1
+rancher_min_version: 2.3.4-rc1
```
cbron: Guessing this will become interactive or not hard coded in future PR ?

cbron (dismissed):
Questions yaml lgtm, that’s important part.

scripts/update-istio-release
daxmc99: Yeah, adding an env var with an echo for now though

scripts/update-istio-release
```diff
@@ -45,11 +45,14 @@ EOF
 done
 rm -r charts/rancher-istio/${1}/charts/istio-init
 
-# Add question.yaml
+# Add question.yaml, make need to update minimum rancher version
+RANCHER_MIN_VERSION="2.3.4-rc1"
+echo "Current rancher min version is set to $RANCHER_MIN_VERSION"
+
```
daxmc99, line 51: @cbron PTAL

cbron (approved):

scripts/update-istio-release
cbron: 👍

PR #151, opened 2020-01-16
rio chart

cbron:
Issue: https://github.com/rancher/rio/issues/339

Few questions to address still with this PR: 
* Is NS label `rio.cattle.io/is-system=true` necessary ? 
* Right now the config only specifies features if they are disabled. Should we put all features in config whether they are true or false ? Guessing this doesn't matter.

### DIFF
--- a/charts/rancher-rio/v0.1.0/templates/clusterrole.yaml
+++ b/charts/rancher-rio/v0.1.0/templates/clusterrole.yaml
@@ -1,0 +1,382 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: rio-cluster-admin
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - '*'
+  verbs:
+  - '*'
+- nonResourceURLs:
+  - '*'
+  verbs:
+  - '*'
+- apiGroups:
+  - apiregistration.k8s.io
+  resources:
+  - apiservices
+  verbs:
+  - '*'
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - mutatingwebhookconfigurations
+  - validatingwebhookconfigurations
+  verbs:
+  - '*'
+- apiGroups:
+  - extensions
+  - apps
+  resources:
+  - replicasets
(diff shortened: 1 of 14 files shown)


StrongMonkey:
> Is NS label rio.cattle.io/is-system=true necessary?

`rio.cattle.io/is-system=true` is used by api-validator webhook.

> Right now the config only specifies features if they are disabled. Should we put all features in config whether they are true or false ? Guessing this doesn't matter.

I think we should specifically put all features in there.

cbron:
Label will have to be added at NS create time. I looked for way to add with Helm but its against best practices. 

> I think we should specifically put all features in there.

This has been updated. 

Also because of what we learned dealing with issues around istio chart versioning, this chart is now not tied to rio version.

charts/rancher-rio/v0.1.0/templates/clusterrole.yaml
```diff
@@ -0,0 +1,382 @@
...
+      - cert-manager.io
+    resources:
+      - '*'
+    verbs:
+      - '*'
+  - apiGroups:
+      - networking.k8s.io
+      - extensions
+    resources:
+      - ingresses
```
StrongMonkey: Something weird with indent...should the array items be align with parent key?
Something like:
```
foo:
- bar1
- bar2
```

cbron: Fixed, good catch.

StrongMonkey (approved):

rancher-max:
Should values.yaml defaults be consistent with rio install today (disable autoscaling, build, and dashboard)?

cbron:
Also need to add cpu/memory requests in order to show UI note if user does not have enough resources: https://github.com/rancher/system-charts/blob/dev/charts/rancher-istio/1.4.300/charts/gateways/values.yaml#L38-L41

daxmc99 (approved):

deniseschannon:
Helm chart for Rio was moved to new format and is located [here](https://github.com/rancher/charts/tree/gh-pages/packages/rio/charts)

PR #152, opened 2020-01-20
Istio 1.4.300

cbron:
Issue: https://github.com/rancher/rancher/issues/25028

* Fix Istio airgap installs
* Fix incorrect image path for sds gateway


superseb:
Can we also remove https://github.com/rancher/system-charts/blob/dev/charts/rancher-istio/1.4.3/charts/grafana/values.yaml#L6-L8 as this causes an additional image to be added to `rancher-images.txt` which is not used.

cbron:
@superseb that will be fixed now. It was incorrectly using `grafana/grafana` instead of our mirror image at `rancher/grafana-grafana`. Thats been updated in values.yaml.

deniseschannon:
@cbron 

The istio chart 2.3.4 needs a min version/max version of 2.3.4, so it's not a 2.3.4 option for any new installs and the new chart we introduce should have a min version of 2.3.4.

rancher-max:
Also address Issue: https://github.com/rancher/rancher/issues/25051

charts/rancher-istio/1.4.3/questions.yaml
```diff
@@ -1,3 +1,4 @@
 labels:
   rancher.istio.v1.4.3: 1.4.3
 rancher_min_version: 2.3.4-rc1
+rancher_max_version: 2.3.4
```
deniseschannon: looking at this, let's match this with the rancher_min_version to 2.3.4-rc1

charts/rancher-istio/1.4.300/questions.yaml
```diff
@@ -1,3 +1,3 @@
 labels:
-  rancher.istio.v1.4.3: 1.4.3
-rancher_min_version: 2.3.4-rc1
+  rancher.istio.v1.4.300: 1.4.3
+rancher_min_version: 2.3.4
```
deniseschannon: let's set this for 2.3.4-rc1 so that anyone deploying a RC would get this chart version.

deniseschannon (changes requested):

charts/rancher-istio/1.4.3/questions.yaml
cbron: Will do. But will it mess with current installs of 1.4.3 on 2.3.4 ? Thought it showed an error about being invalid version or something.

charts/rancher-istio/1.4.300/questions.yaml
cbron: fixed

deniseschannon (approved):
I mostly reviewed for formatting on how a system-chart should work in Rancher, not for functionality of Istio.

PR #157, opened 2020-02-13
Istio 1.4.400

cbron:
Issue: https://github.com/rancher/rancher/issues/25403

Working off the release-v2.3 branch.

Depends on images from linked rancher PR in issue.


charts/rancher-istio/1.4.400/requirements.yaml
```diff
@@ -1,40 +1,40 @@
...
-    version: 1.4.3
+    version: 1.4.4
     condition: tracing.enabled
   - name: galley
-    version: 1.4.3
+    version: 1.4.4
     condition: galley.enabled
   - name: kiali
-    version: 1.4.3
+    version: 1.9.0
```
cbron: Kiali chart doesn't use current istio version

StrongMonkey: hmm, this version doesn't mean image version, correct? It means chart version.

cbron: Correct, although the image doesn't match either. Its pointing at sub-chart version, which wasn't matching on 1.4.300 either: https://github.com/rancher/system-charts/blob/dev/charts/rancher-istio/1.4.300/charts/kiali/Chart.yaml#L7

cbron: reverted to keep in line with upstream

StrongMonkey (dismissed):
LGTM, would prefer this to be split into two PRs, one for rebase and one for istio 1.4.3.

cbron:
Yes, up to @deniseschannon how we should rebase. This was off release-v2.3, but I can cherry pick to wherever.

deniseschannon (approved):

PR #162, opened 2020-02-19
OPA Gatekeeper system chart

prachidamle:
- Adding basecopy of the OPA Gatekeeper helm v2 chart
- Adding builtin ConstraintTemplates
- Adding nodeSelector to ensure the workloads never schedule to the Windows node
- Add private image registry for airgap case

https://github.com/rancher/rancher/issues/23753

Related PR: https://github.com/rancher/rancher/pull/25578


charts/rancher-gatekeeper-operator/v3.1.0-beta.700/values.yaml
```diff
@@ -6,7 +6,7 @@ image:
   repository: quay.io/open-policy-agent/gatekeeper
```
StrongMonkey: Wasn't quite sure but looks like we have to a pattern to use images under `rancher` repo. Can we add this to our mirror scripts https://github.com/rancher/rancher/blob/master/scripts/istio-images-mirror?

charts/rancher-gatekeeper-operator/v3.1.0-beta.700/templates/gatekeeper.yaml
```diff
@@ -485,7 +485,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
-        image: '{{ .Values.image.repository }}:{{ .Values.image.release }}'
+        image: {{ template "system_default_registry" . }}'{{ .Values.image.repository }}:{{ .Values.image.release }}'
```
cbron: Seems odd to have just the registry outside of the quotes, typically its either all inside or all outside. Although Helm probably doesn't care.

charts/rancher-gatekeeper-operator/v3.1.0-beta.700/values.yaml
cbron: Yes if we want to do airgap, in order to get them included in rancher-images.txt (which people use as a list of images to download) then all images must be in format of:
```
repository: user/image
tag: 1.2.3
```
You can test this from rancher by running `go run main.go ~/path/to/system-charts` from rancher/pkg/image/export/main.go. You'll see that this one wasn't picked up, or at least its tag wasn't. The istio-images-mirror script is what gets them into our dockerhub.

deniseschannon:
https://github.com/rancher/rancher/issues/23753

charts/rancher-gatekeeper-operator/v3.1.0-beta.700/values.yaml
prachidamle: Added to the script https://github.com/rancher/rancher/pull/25578

cbron (dismissed):
LGTM. Would be good for QA to do airgap testing.

StrongMonkey (dismissed):

charts/rancher-gatekeeper-operator/v3.1.0-beta.700/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+description: A Helm chart for Gatekeeper
+name: gatekeeper-operator
```
deniseschannon: Since all the other charts are named `rancher-X`, should we update this chart to be consistent?

charts/rancher-gatekeeper-operator/v3.1.0-beta.700/Chart.yaml
```diff
@@ -0,0 +1,10 @@
...
+keywords:
+  - open policy agent
+version: v3.1.0-beta.7
```
deniseschannon: We should consider adding `00` to the end of this in case we need to make changes to the chart before the app version changes.

deniseschannon (changes requested):

charts/rancher-gatekeeper-operator/v3.1.0-beta.700/Chart.yaml
cbron: Depending on how this is implemented in the UI, we might need to add a [questions.yaml](https://github.com/rancher/system-charts/blob/dev/charts/rancher-istio/1.4.400/questions.yaml#L2) to label this correctly. Apparently doesn't work with monitoring because of how it's implemented in the UI, so I'm not sure about this case.

charts/rancher-gatekeeper-operator/v3.1.0-beta.700/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+description: A Helm chart for Gatekeeper
+name: rancher-gatekeeper-operator
+keywords:
+  - open policy agent
+version: v3.1.0-beta.700
```
deniseschannon: Based on the strict helm semver comparisons, I wonder if we should start with v0.1.0 for this helm chart until it actually GA's and then use the official version from Gatekeeper.

charts/rancher-gatekeeper-operator/v3.1.0-beta.700/helm-modifications/helm-modifications.yaml
```diff
@@ -0,0 +1,61 @@
...
+        - name: manager
+          args:
+            - --audit-interval={{ .Values.auditInterval }}
+            - --port=8443
+            - --logtostderr
+            - --constraint-violations-limit={{ .Values.constraintViolationsLimit }}
+            - --audit-from-cache={{ .Values.auditFromCache }}
+            - --exempt-namespace=gatekeeper-system
+          imagePullPolicy: "{{ .Values.image.pullPolicy }}"
+          image: "{{ .Values.image.repository }}:{{ .Values.image.release }}"
```
deniseschannon: why is this `Values.image.release`? Shouldn't it be `Values.image.tag`?

deniseschannon: Also, it seems to be missing logic to handle the default system registry for when you will use airgap and have your own private registry: https://github.com/rancher/system-charts/blob/dev/charts/rancher-istio/1.4.400/charts/istiocoredns/templates/deployment.yaml#L40

deniseschannon (changes requested):

charts/rancher-gatekeeper-operator/v3.1.0-beta.700/helm-modifications/helm-modifications.yaml
prachidamle: This yaml file does not seem to be used, airgap and the Values.image.tag has been taken care of in gatekeeper.yaml @deniseschannon

deniseschannon:
Can we also add in a min Rancher version of 2.4.0-rc1?

charts/rancher-gatekeeper-operator/v3.1.0-beta.700/helm-modifications/helm-modifications.yaml
deniseschannon: Do we need this file then?

prachidamle: Dint want to remove anything from the source copy of the chart copied from gatekeeper project. not sure if we should remove, I can test once without it and check

charts/rancher-gatekeeper-operator/v3.1.0-beta.700/Chart.yaml
prachidamle: not sure about the reason, will discuss this

deniseschannon (approved):

PR #164, opened 2020-02-21
System upgrade controller

rmweir:


daxmc99:
What is the purpose of the path `charts/system-upgrade-controller/v0.3.0-m4/Chart.yaml`? Should we use just `v0.3.0` or does `-m4` carry significance?

daxmc99 (commented):
What is the purpose of the path `charts/system-upgrade-controller/v0.3.0-m4/Chart.yaml`? Should we use just `v0.3.0` or does `-m4` carry significance?

rmweir:
> What is the purpose of the path `charts/system-upgrade-controller/v0.3.0-m4/Chart.yaml`? Should we use just `v0.3.0` or does `-m4` carry significance?

it the version of the image that is being used by default. It happens to be the newest release for SUC and there is no v0.3.0 yet. Which brings me to another question- currently I have the version as configurable so really the chart could be for any future version of SUC that can be used with the reset of the chart. I'm wondering if I should that configurability out? If not what should the path actually be?

dweomer (dismissed):
@rmweir LGTM 👍

charts/system-upgrade-controller/v0.3.0-m4/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+name: system-upgrade-controller
```
deniseschannon: The naming gods have declared `rancher-k3s-upgrader`

charts/system-upgrade-controller/v0.3.0-m4/templates/deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+              - matchExpressions:
+                  - {key: "node-role.kubernetes.io/master", operator: In, values: ["true"]}
+      serviceAccountName: system-upgrade
+      tolerations:
+        - key: "node.kubernetes.io/unschedulable" # necessary to avoid cordon/drain
+          operator: "Exists"
+          effect: "NoSchedule"
+      containers:
+        - name: system-upgrade-controller
+          image: rancher/system-upgrade-controller:v0.3.0-m4
```
deniseschannon: Where/how is this image in Dockerhub? Also, what happens when/if someone is in airgap and wants to use their private registry?

deniseschannon (changes requested):
Can we also put a rancher min version of v2.4 in a questions.yml?

charts/system-upgrade-controller/v0.3.0-m4/Chart.yaml
```diff
@@ -0,0 +1,8 @@
...
+description: Enables a k3s cluster to update itself by reacting to Plan CRs.
+home: https://github.com/rancher/system-charts/charts/system-upgrade-controller
+sources:
+  - "https://github.com/rancher/system-charts/charts/system-upgrade-controller"
+version: 0.1.0
+appVersion: "1.16.0"
```
deniseschannon: Why is this set to 1.16.0?

deniseschannon (changes requested):

charts/rancher-k3s-upgrader/v0.0.1/templates/deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+              - matchExpressions:
+                  - {key: "node-role.kubernetes.io/master", operator: In, values: ["true"]}
+      serviceAccountName: system-upgrade
+      tolerations:
+        - key: "node.kubernetes.io/unschedulable" # necessary to avoid cordon/drain
+          operator: "Exists"
+          effect: "NoSchedule"
+      containers:
+        - name: system-upgrade-controller
+          image: rancher/system-upgrade-controller:v0.3.0-rc1
```
deniseschannon: This can't be hardcoded as it will never get updated if you have a default registry from Rancher (for airgap setups) or tags? 

I expected something like this: https://github.com/rancher/system-charts/blob/dev/charts/rancher-istio/1.4.400/charts/istiocoredns/templates/deployment.yaml#L40

deniseschannon (changes requested):

deniseschannon (approved):

PR #173, opened 2020-03-05
Monitoring metrics

mrajashree:


deniseschannon:
v0.0.8 is the new chart for 2.4, so we shouldn't be creating a new chart since it hasn't been released.

PR #179, opened 2020-03-11
Add rancher-logging 0.1.5

superseb:
https://github.com/rancher/rancher/issues/25182

This broke because privileged was removed in https://github.com/rancher/system-charts/pull/89/.

This adds the option to provide `global.privileged = true` to add `privileged: true` to the launched pods for Rancher logging. This is a post upgrade action on the App created in the System project. This choice was made versus the choice to make this the default, as it adds privileged to the pods and it's only needed for a small percentage of installations (RedHat/CentOS + SELinux enforcing + Docker selinux enabled). This needs to be added to the docs.

The choice here is to make it work for everyone (by adding privileged by default) or make it an option for the smaller set of usecases (see https://github.com/fluent/fluentd-kubernetes-daemonset#running-on-openshift as an example as well)

### DIFF
--- a/charts/rancher-logging/0.1.5/questions.yml
+++ b/charts/rancher-logging/0.1.5/questions.yml
@@ -1,0 +1,1 @@
+rancher_min_version: 2.3.4-rc1
--- a/charts/rancher-logging/0.1.5/.helmignore
+++ b/charts/rancher-logging/0.1.5/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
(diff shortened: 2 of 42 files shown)


charts/rancher-logging/0.1.5/questions.yml
```diff
@@ -1 +1 @@
-rancher_min_version: 2.3.4-rc1
\ No newline at end of file
+rancher_min_version: 2.4.0-rc1
```
deniseschannon: Do we really need the min version for 2.4.0?

deniseschannon:
In the app readme, add in a change log of what's been updated.

charts/rancher-logging/0.1.5/questions.yml
deniseschannon: I'd like to keep the Rancher min version the same and add in comments into the `app-readme.md` of what's the changelog.

deniseschannon (approved):

PR #182, opened 2020-03-13
Adding a pre-delete hook - job to delete constrainttemplate crds

prachidamle:
https://github.com/rancher/rancher/pull/25575

Adding in a pre-delete hook to the helm chart to clean up OPA Gatekeeper constrainttemplates during removal of the helm release.

### DIFF
--- a/charts/rancher-gatekeeper-operator/v0.1.0/templates/job-constraints-crd.yaml
+++ b/charts/rancher-gatekeeper-operator/v0.1.0/templates/job-constraints-crd.yaml
@@ -1,0 +1,19 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: gatekeeper-delete-constraints-crd-job
+  annotations:
+    "helm.sh/hook": "pre-delete"
+    "helm.sh/hook-delete-policy": "hook-succeeded, before-hook-creation, hook-failed"
+spec:
+  template:
+    spec:
+      serviceAccountName: gatekeeper-admin
+      containers:
+      - name: gatekeeper-delete-constraints-crd
+        image: "{{ template "system_default_registry" . }}{{ .Values.global.kubectl.repository }}:{{ .Values.global.kubectl.tag }}"
+        imagePullPolicy: {{ .Values.global.imagePullPolicy }}
+        command: ["kubectl",  "delete", "constrainttemplates", "--all", "--wait=false"]
+      restartPolicy: Never
+  backoffLimit: 1
--- a/charts/rancher-gatekeeper-operator/v0.1.0/values.yaml
+++ b/charts/rancher-gatekeeper-operator/v0.1.0/values.yaml
@@ -17,3 +17,6 @@ resources:
     memory: 256Mi
 global:
   systemDefaultRegistry: ""
+  kubectl:
+    repository: rancher/istio-kubectl
+    tag: 1.4.6


luthermonson (approved):

charts/rancher-gatekeeper-operator/v0.1.0/templates/job-constraints-crd.yaml
```diff
@@ -0,0 +1,19 @@
...
+    "helm.sh/hook-delete-policy": "hook-succeeded, before-hook-creation, hook-failed"
+spec:
+  template:
+    spec:
+      serviceAccountName: gatekeeper-admin
+      containers:
+      - name: gatekeeper-delete-constraints-crd
+        image: "{{ template "system_default_registry" . }}{{ .Values.global.kubectl.repository }}:{{ .Values.global.kubectl.tag }}"
+        imagePullPolicy: {{ .Values.global.imagePullPolicy }}
+        command: ["kubectl",  "delete", "constrainttemplates", "--all", "--wait=false"]
```
StrongMonkey, line 17: Just to make sure, does the crd have finalizers? if so does it prevent crd from being deleted?

prachidamle: No the CRD has no finalizers.

StrongMonkey: :+1:

StrongMonkey (approved):

charts/rancher-gatekeeper-operator/v0.1.0/templates/job-constraints-crd.yaml
cjellick: Is your intent to fully uninstall the CRDs or just the instances of the CRDs?

If your intent is to follow the instructions from the OPA documentation, you aren't fully accomplishing that.
https://github.com/open-policy-agent/gatekeeper#manually-removing-constraints
```
If Gatekeeper is no longer running and there are extra constraints in the cluster, then the finalizers, CRDs and other artifacts must be removed manually:

Delete all instances of the constraint resource
Executing kubectl patch crd constrainttemplates.templates.gatekeeper.sh -p '{"metadata":{"finalizers":[]}}' --type=merge. Note that this will remove all finalizers on every CRD. If this is not something you want to do, the finalizers must be removed individually.
Delete the CRD and ConstraintTemplate resources associated with the unwanted constraint.
```

You are just deleting the CRs, not the CRD. These instructions say to delete the CRD itself.

Moreover, you aren't accounting for the fact that if these resources have finalizers and the controller in charge of removing the finalizer is somehow sideways or not running, the CRs will get wedged in a deleting state.

prachidamle: @cjellick  The above instructions you posted are to be followed in the case when gatekeeper is already uninstalled and no longer running. In that case any finalizers that are lingering around (which gatekeeper could have cleared),  need to be removed manually by `Executing kubectl patch crd ... `

What I am doing is following this https://github.com/open-policy-agent/gatekeeper#uninstallation 

```Before uninstalling Gatekeeper, be sure to clean up old Constraints, ConstraintTemplates, and the Config resource in the gatekeeper-system namespace. This will make sure all finalizers are removed by Gatekeeper. Otherwise the finalizers will need to be removed manually.```

Specifically, I am cleaning up of resources (which cleans up the CRDs inturn) while gatekeeper is still running - using pre-delete hook. The instructions are here: https://github.com/open-policy-agent/gatekeeper#before-uninstall-clean-up-old-constraints

```When **Gatekeeper is running** it is possible to remove unwanted constraints by:

Deleting all instances of the constraint resource
Deleting the ConstraintTemplate resource, which should automatically clean up the CRD
Deleting the Config resource removes finalizers on synced resources
```

In our current usage of gatekeeper, we are not creating any Config resource, so those we need not cleanup. And in my tests I see that deleting ConstraintTemplates will also delete the Constraints created of those type.

prachidamle: Given above, the whole intent of this is only to unblock the "Disable Gatekeeper" usecase for Rancher. Currently if these constrainttemplate resources and their CRDs are not cleaned up when gatekeeper is deleted, one cannot re-install the gatekeeper app.
After clearing up these resources, reinstallation is not blocked.

prachidamle: >>>Moreover, you aren't accounting for the fact that if these resources have finalizers and the controller in charge of removing the finalizer is somehow sideways or not running, the CRs will get wedged in a deleting state.

@cjellick  This is the only thing that I am not completely sure about - If the finalizers are owned by gatekeeper - they will be taken care of by the running gatekeeper - thats what the documentation says about deleting constraints while gatekeeper is still running.

But if some custom finalizer is added by the user that may block us if the controller in charge is off.
Maybe clearing out all finalizers before delete might save us in that case, would it?

cjellick: @prachidamle here's my reasoning regarding the finalizers:
First, i'm not worried about finalizers added by some random third party. That is unlikely. What I am worried about is situations where the user is uninstalling gatekeeper because something bad has happened and gatekeeper is not functioning properly.

One possible scenario (just an example, not sure if it could actually occur, that would depend on a lot of factors): Imagine if your cluster had a problem where there are no schedulable nodes. If the gatekeeper wasnt already running (like if the node it was on shut down), it would not get launched. If you then try to uninstall gatekeeper, you'll get resources stuck in deleting bc of the finalizers.

On a lot apps, we could more or less live with this, but Gatekeeper has a significant impact on a cluster. It can reject requests, so if things go badly with it, it could be pretty impactful. So, I'd prefer to be agressive in the cleanup.

cjellick: On the other hand, we are possibly prematurely deleting the finalizer without letting the controller do it. That could have an impact. So...I think you don’t need to delete the finalizers here.

Next Question: do you need to delete the actual CRD for reinstall to work or just the CRs?

prachidamle: @cjellick  correct, gatekeeper should be removing finalizers, that is the design of cleanup before uninstalling gatekeeper, as documented.

I don't need to delete the actual CRD for reinstall - they get removed when CRs are removed, after this pre-delete hook runs. This is also documented and I see that in my tests.

https://github.com/open-policy-agent/gatekeeper#before-uninstall-clean-up-old-constraints
```Deleting the ConstraintTemplate resource, which should automatically clean up the CRD```

prachidamle: This is the kubectl output while gatekeeper is running:

Screen Shot 2020-03-15 at 12 50 58 PM

After removing gatekeeper app from Rancher, this is the kubectl output showing CRDs are also removed with constrainttemplates
Screen Shot 2020-03-15 at 12 52 29 PM

prachidamle: The base CRDs that gatekeeper's chart adds, namely `constrainttemplates.templates.gatekeeper.sh` and `configs.config.gatekeeper.sh` dont prohibit re-installation.

cjellick (approved):
Ok, looks good

PR #185, opened 2020-03-16
Recording rule for apiserver metrics

mrajashree:
The metric `apiserver_request_count` was renamed to `apiserver_request_total`, and the deprecated `apiserver_request_count` was removed from k8s 1.17
This PR adds recording rule to rename `apiserver_request_count` to `apiserver_request_total` for clusters with k8s version 1.17 and up

https://github.com/kubernetes/kubernetes/pull/72336
https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/metrics/metrics.go


daxmc99 (dismissed):
This has the impact of allowing `apiserver_request_count` (which was deprecated in 1.17) to be collected as `apiserver_request_total`

daxmc99 (approved):

deniseschannon:
Closing as it was brought in with #226

PR #190, opened 2020-03-20
Add tolerations in values.yaml

StrongMonkey:


charts/rancher-cis-benchmark/v0.0.1/values.yaml
```diff
@@ -66,7 +68,8 @@ resources: {}
 
 nodeSelector: {}
 
-tolerations: []
+tolerations: 
+- operator: Exists
```
prachidamle, line 72: Why are we changing the kube-bench tolerations to match all @StrongMonkey?

StrongMonkey: We need to be able to schedule this pod to any host. In a setup with only etcd and controlplane nodes, CIS can't be run.

prachidamle (approved):
LGTM

leodotcloud (approved):
LGTM

PR #193, opened 2020-03-25
Add experimental label to OPA gatekeeper chart

prachidamle:
@deniseschannon forgot about this; not sure if this is the correct way to add this though.

@westlywright will UI be able to consume this and show an "Experimental" tag on Gatekeeper Page?

### DIFF
--- a/charts/rancher-gatekeeper-operator/v0.1.0/questions.yaml
+++ b/charts/rancher-gatekeeper-operator/v0.1.0/questions.yaml
@@ -1,1 +1,3 @@
 rancher_min_version: 2.4.0-rc1
+labels:
+  io.rancher.certified: experimental


charts/rancher-gatekeeper-operator/v0.1.0/questions.yaml
```diff
@@ -1 +1,2 @@
-rancher_min_version: 2.4.0-rc1
\ No newline at end of file
+rancher_min_version: 2.4.0-rc1
+io.rancher.certified: experimental
```
deniseschannon: this needs to be nested under labels @prachidamle

deniseschannon (changes requested):

charts/rancher-gatekeeper-operator/v0.1.0/questions.yaml
prachidamle: @deniseschannon rancher_min_version as well?

deniseschannon (approved):

PR #198, opened 2020-03-30
Istio 1.4.700 - 2.3

cbron:
rancher/rancher#26232


cbron:
@deniseschannon This is against release-2.3, other 2 are against dev-2.X, that right ?

deniseschannon:
@cbron No, this should be against `dev-v2.3` for it to be tested before we push it out.

deniseschannon (approved):

PR #205, opened 2020-04-14
OPA Gatekeeper update chart to use new beta image v3.1.0-beta.8

prachidamle:
https://github.com/rancher/rancher/issues/26453

### DIFF
--- a/charts/rancher-gatekeeper-operator/0.1.1/Chart.yaml
+++ b/charts/rancher-gatekeeper-operator/0.1.1/Chart.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+description: A Helm chart for Gatekeeper
+name: rancher-gatekeeper-operator
+keywords:
+  - open policy agent
+version: 0.1.0
+home: https://github.com/open-policy-agent/gatekeeper
+sources:
+  - https://github.com/open-policy-agent/gatekeeper.git
+appVersion: v3.1.0-beta.7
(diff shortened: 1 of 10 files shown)


charts/rancher-gatekeeper-operator/0.1.1/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+description: A Helm chart for Gatekeeper
+name: rancher-gatekeeper-operator
+keywords:
+  - open policy agent
+version: 0.1.0
```
deniseschannon: This should be updated to 0.1.1 to match the version

charts/rancher-gatekeeper-operator/0.1.1/Chart.yaml
```diff
@@ -0,0 +1,10 @@
...
+home: https://github.com/open-policy-agent/gatekeeper
+sources:
+  - https://github.com/open-policy-agent/gatekeeper.git
+appVersion: v3.1.0-beta.7
```
deniseschannon: This should also be updated to the new image to match.

prachidamle: done

deniseschannon:
We should eventually add in `rancher_max_version:` to the old chart, but would like to see what hte experience of upgrading is like in the new workflow.

deniseschannon (approved):

PR #206, opened 2020-04-15
Change to use the rancher generated image instead of upstream

prachidamle:
We need to use the rancher generated image instead of the upstream image directly. The corresponding rancher PR to generate the image is already merged to master: 
https://github.com/rancher/rancher/pull/26517


dramich:
@deniseschannon Does this look OK to you?

deniseschannon (approved):

PR #208, opened 2020-04-24
Istio 1.4.800 - 2.5

brendarearden:
https://github.com/rancher/rancher/issues/26784


daxmc99 (dismissed):

charts/rancher-istio/1.4.800/questions.yaml
```diff
@@ -1,3 +1,3 @@
 labels:
-  rancher.istio.v1.4.700: 1.4.7
+  rancher.istio.v1.4.800: 1.4.8
 rancher_min_version: 2.3.4-rc1
```
cbron, line 3: Need to ask if we should also deprecate the 1.4.700 chart, think the answer is no.

cbron (commented):
Just a question but looks good. After this is merged, need to also PR to dev-v2.3 and dev-v2.4 branches.

charts/rancher-istio/1.4.800/questions.yaml
brendarearden: per Denise we will be deprecating, so I will update the pr

charts/rancher-istio/1.4.700/questions.yaml
```diff
@@ -1,3 +1,4 @@
 labels:
   rancher.istio.v1.4.700: 1.4.7
 rancher_min_version: 2.3.4-rc1
+rancher_max_version: 2.3.4-rc2
```
cbron: You can keep this as rc1

charts/rancher-istio/1.4.800/questions.yaml
```diff
@@ -1,3 +1,3 @@
 labels:
   rancher.istio.v1.4.800: 1.4.8
-rancher_min_version: 2.3.4-rc1
+rancher_min_version: 2.3.4-rc2
```
cbron: And set min to rc1 here

cbron (changes requested):
Small change, doesn't really matter but want to keep simple as not to be confusing in future for others.

cbron (approved):
lgtm

PR #226, opened 2020-05-13
Monitoring v0.1.1

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rancher/issues/26969

StrongMonkey:
Include https://github.com/rancher/system-charts/pull/185

guangbochen (commented):
LGTM of the template changes, since we have bumped the prometheus-operator images, a very important thing to verify is that the CRDs are also updated to the newer version of existing deployment, otherwise prometheus-operator may encounter some incompatible error due to CRD changes.

thxCode (approved):
LGTM, agree with @guangbochen.

PR #227, opened 2020-05-14
rancher-monitoring: support prometheus/alertmanager-instance-namespaces

iwilltry42:
## What?

This PR introduces a new value `denyNamespaces` for the operator.
If this is specified, it will add the `--deny-namespaces=<list>` arg on the prometheus-operator.

## Why?

This is especially useful if someone tries to use another prometheus-operator on the same cluster.
E.g. I deployed the prometheus-operator helm chart to the namespace `prometheus-operator` (operator v0.38.1 and v0.39.0 testes) and the alertmanager and prometheus server failed to start, because they were constantly being updated by the operator.
Interestingly, this stopped when I added the `--deny-namespaces=prometheus-operator` arg to the operator in `cattle-prometheus` or when I deleted the `cattle-prometheus` namespace (which will automatically be recreated).
So I guess that the Rancher prometheus-operator and the custom deployed operator are fighting to update the resources, which then results in an update/termination loop.

### DIFF
--- a/charts/rancher-monitoring/v0.1.0/Chart.yaml
+++ b/charts/rancher-monitoring/v0.1.0/Chart.yaml
@@ -7,7 +7,7 @@ maintainers:
 name: rancher-monitoring
 sources:
 - https://github.com/coreos/prometheus-operator
-version: 0.1.0
+version: 0.1.1
 appVersion: 0.1.0
 home: https://github.com/coreos/prometheus-operator
 keywords:
--- a/charts/rancher-monitoring/v0.1.0/charts/operator/templates/deployment.yaml
+++ b/charts/rancher-monitoring/v0.1.0/charts/operator/templates/deployment.yaml
@@ -51,6 +51,12 @@ spec:
             - --crd-apigroup={{ template "operator_api_group" . }}
             - --manage-crds={{ .Values.manageCRDs }}
             - --with-validation={{ .Values.withValidation }}
+            {{- if .Values.prometheusInstanceNamespaces }}
+            - --prometheus-instance-namespaces={{ .Values.prometheusInstanceNamespaces | join "," }}
+            {{- end }}
+            {{- if .Values.alertmanagerInstanceNamespaces }}
+            - --alertmanager-instance-namespaces={{ .Values.alertmanagerInstanceNamespaces | join "," }}
+            {{- end }}
           ports:
           - containerPort: 8080
             name: http
--- a/charts/rancher-monitoring/v0.1.0/values.yaml
+++ b/charts/rancher-monitoring/v0.1.0/values.yaml
@@ -27,6 +27,8 @@ operator:
   ## Already exist ServiceAccount
   ##
   serviceAccountName: ""
+  prometheusInstanceNamespaces: []
+  alertmanagerInstanceNamespaces: []
 
 exporter-coredns:
   enabled: false


iwilltry42:
This could probably be merged with #226

StrongMonkey:
@iwilltry42 Thanks for adding this :) @thxCode @guangbochen can you help review this PR? It looks good to me, the option is already available in upstream chart.

iwilltry42:
There are surely more useful options in the upstream chart, but without this specific one, extra operator deployments fail and one has to add this to the deployment manually. 👍

thxCode:
In my understanding of the introduction of this parameter: `--deny-namespaces`, this is not intended to deal with such scenarios. It is a parameter that is mutually exclusive with `--namespaces`.

For two or more Prometheus Operators which have different purposes in the cluster, this is really difficult to solve through Rancher. Through `deny-namespaces`, it will only make the entire management more opaque.

@iwilltry42 , I think you should analyze the actual cause of the flipping altermanager and prometheus server in the your scenarios. Is the configuration of the operator causing a label conflict in the generated workload or conflict in executing the command?

cc @StrongMonkey

iwilltry42:
Hi @thxCode, you're probably right in a way that the `--deny-namespaces` parameter is not the correct fix for my use case.
I should probably update the PR to rather introduce the `--{prometheus, alertmanager}-instance-namespaces` flags instead, which only describe the list of namespaces where Prometheus and Alertmanager instances will be managed (see https://github.com/coreos/prometheus-operator/blob/c7405c847b636e174ce7c2306c57619bca8fcee9/cmd/operator/main.go#L177-L178 & https://github.com/coreos/prometheus-operator/pull/2783).
This also takes precedence over the `--namespaces` and `--deny-namespaces` parameters.

One reason for the issues could be the updated volumeClaimTemplate stuff (see https://github.com/coreos/prometheus-operator/issues/2950) and the difference in versions used for the prometheus-operator (i.e. version mismatch between Rancher's and my custom prometheus-operator deplyoment).

In any case, I don't think that I should have to adapt my custom prometheus-operator deployment to the built-in operator, which I cannot modify enough to use it properly and which I also cannot delete entirely :thinking: 

UPDATE: I pushed a new commit to switch to the other flags. However, I think one could have default values here to ensure that at least (or only) the `cattle-prometheus` namespace is included.

thxCode (dismissed):
LGTM

iwilltry42:
Should the following be the default here?
```yaml
prometheusInstanceNamespaces:
  - cattle-prometheus
alertmanagerInstanceNamespaces:
  - cattle-prometheus
```
I have the feeling, that it's not expected to be managing other instances as well? :thinking: 
@thxCode what do you think?

iwilltry42:
@thxCode , @guangbochen , @StrongMonkey is there anything left todo here?
What do you think about limiting the operator to managing only instances in `cattle-prometheus` namespace by default?

codayblue:
I am running into an issue with the portworx prometheus operator. The change in this PR solved it. What is needed to get this merged? If the original author is busy I can correct the issues on the PR and get it reposted.

iwilltry42:
Updated to `dev-v2.6` base branch (even though my fork's branch name is still dev-v2.5, it's on dev-v2.6 :grin: ) :+1: 

@StrongMonkey & @guangbochen , maybe this could be reviewed again?

ggriffiths:
Hello @iwilltry42 @StrongMonkey @guangbochen  I ran across this PR as we were tracking for the issue @codayblue mentioned. Would it be possible to get this reviewed? 

I'm guessing it would maybe need to be rebased against a newer version? I can assist with this if needed.

recena:

PR #228, opened 2020-05-20
Istio 1.4.900

brendarearden:
Adding Istio version 1.4.9

**Issue**
https://github.com/rancher/rancher/issues/27064


brendarearden:
https://github.com/istio/istio/compare/release-1.4.7-patch...release-1.4.9-patch

charts/rancher-istio/1.4.900/questions.yaml
```diff
@@ -1,4 +1,3 @@
 labels:
-  rancher.istio.v1.4.700: 1.4.7
-rancher_min_version: 2.3.4-rc1
-rancher_max_version: 2.3.6
+  rancher.istio.v1.4.900: 1.4.9
+rancher_min_version: 2.3.7-rc1
```
cbron, line 3: Min version should be 2.3.4-rc1, as we want 1.4.9 to surpass 1.4.7.

cbron: Also 1.4.700 should max at same point, same as 1.4.600 is now: https://github.com/rancher/system-charts/blob/dev-v2.5/charts/rancher-istio/1.4.600/questions.yaml

cbron (changes requested):
Looks like 1.4.8 changes got incorporated so good there, just need to adjust questions.

charts/rancher-istio/1.4.900/questions.yaml
brendarearden: Changed min version to 2.3.4-rc1. I had missed this requirement in the issue, so now the PR meets all requirements.

cbron (dismissed):
lgtm, but would like Denise approval

charts/rancher-istio/1.4.900/questions.yaml
cbron: @deniseschannon want to confirm this ^, as this is a security release we want to go all the way back. Istio 1.4.8 only went back to 2.3.7 because it was non-security patch release.

charts/rancher-istio/1.4.700/questions.yaml
```diff
@@ -1,4 +1,4 @@
 labels:
   rancher.istio.v1.4.700: 1.4.7
 rancher_min_version: 2.3.4-rc1
-rancher_max_version: 2.3.6
+rancher_max_version: 2.3.4-rc1
```
deniseschannon: let's not change the max version and keep this as up to 2.3.6.

charts/rancher-istio/1.4.800/questions.yaml
```diff
@@ -1,3 +0,0 @@
-labels:
-  rancher.istio.v1.4.800: 1.4.8
-rancher_min_version: 2.3.7-rc1
```
deniseschannon, line 3: Why was this deleted? 

Let's keep the min version as is for 2.3.7-rc1.

deniseschannon (changes requested):
Let's keep 1.4.700 as the version shipped with 2.3.4-2.3.6 and make 1.4.900 what's available from 2.3.7+

charts/rancher-istio/1.4.800/questions.yaml
cbron: We aren't ever deploying 1.4.8, 1.4.7 is last version from 1.4.9

cbron (approved):
Lgtm, 1.4.7 is already maxed at 2.3.6.

deniseschannon (approved):

PR #229, opened 2020-05-20
Add thano rules into rbac

StrongMonkey:


StrongMonkey:
https://github.com/helm/charts/commit/347d7a627be3a695c6500926bad888e9887a2c66#diff-3aa210e2fed854e7e981213b8dce54a0  https://github.com/rancher/rancher/issues/26969#issuecomment-629565133

StrongMonkey:
cc @ekristen

ekristen:
Are these needed? I'm not clear why they are in the prometheus operator. The thanos rulers are suppose to be created on the thanos-ruler service to be executed by thanos-ruler against thanos-query and stored back against the thanos-store.

I don't think it hurts anything being here, but not sure the motivation by prometheus operator to include them when prometheus operator doesn't do a thanos deployment just the sidecar deployment.

StrongMonkey:
Yeah I think it is not necessary since it was working before. But the misleading log on the prometheus-operator seems to confuse users.

deniseschannon (approved):

PR #232, opened 2020-05-22
[2.4] Backport monitoring v0.1.1 into dev-v2.4

StrongMonkey:


charts/rancher-monitoring/v0.1.1/questions.yml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.4.0-rc1
```
deniseschannon, line 1: If we're replacing this as of 2.4.0-rc1, we should make monitoring v0.1.0 with a max version to remove it.

deniseschannon (changes requested):

deniseschannon (approved):

PR #235, opened 2020-05-26
Updating the new version to use the latest new rancher/security-scan tag

prachidamle:
New tag is released for rancher/security-scan image; updating to use this tag.


cbron:
Appears that 0.1.1 has not been released yet, but would we need to bump chart for internal use ?

deniseschannon (approved):

deniseschannon:
@cbron correct, they are dev testing CIS scans, so they are just bumping the image in the `dev-v2.5` branch, which doesn't mean you need to bump the chart folder. 

We only ever want to provide 1 new chart per release.

PR #234, opened 2020-05-22
CIS benchmark: Updating chart to use new  security-scan image

prachidamle:
There is a new image  rancher/security-scan:v0.1.10

This brings in the fixes necessary for CIS 1.5 tests.

PR #236, opened 2020-05-29
[BackwardPort] CIS benchmark: Updating chart to add new version and use new security-scan image 

prachidamle:
Backward port following PRs from branch dev-v2.5:
rancher/system-charts#234
rancher/system-charts#235


cbron (approved):
lgtm

PR #246, opened 2020-06-08
Add selector for exporter-node-windows service

### LINKED ISSUE #27483: Missing selector for exporter-node-windows service
We have a custom cluster (not RKE) created via kubeadm that includes windows nodes and we manage it through rancher.

To enable windows node exporters, we add the following answer under advanced options in cluster monitoring:

    exporter-node-windows.enabled = true

Everything gets deployed fine and the exporters start on the windows hosts.
Prometheus gets configured with new targets and the windows specific rules are set up.

The problem is that prometheus does not find anything under kubernetes service discovery.

From what I understand, in an RKE cluster, new [Endpoints](https://github.com/rancher/system-charts/blob/dev-v2.5/charts/rancher-monitoring/v0.1.0/charts/exporter-node-windows/templates/endpoint.yaml) will be created with a list of IP address.
This is not that convenient, as we don't want to update that list every time a node joins or leaves the cluster.

With some digging, and comparing the [exporter-node](https://github.com/rancher/system-charts/tree/dev-v2.5/charts/rancher-monitoring/v0.1.0/charts/exporter-node) and [exporter-node-windows](https://github.com/rancher/system-charts/tree/dev-v2.5/charts/rancher-monitoring/v0.1.0/charts/exporter-node-windows) system charts I found out that the service.yaml for windows does not have a selector.

By manually editing the service and adding the following labels for the selector, things start working normally:

      app: exporter-node-windows
      chart: exporter-node-windows-0.0.1
      release: cluster-monitoring

Adding something like:

      selector:
        app: {{ template "app.name" . }}
        chart: {{ template "app.version" . }}
        release: {{ .Release.Name }}

under `spec` in [service.yaml](https://github.com/rancher/system-charts/blob/dev-v2.5/charts/rancher-monitoring/v0.1.0/charts/exporter-node-windows/templates/service.yaml) will work for our use case.

Are there any specific reasons why the selector is missing?

---
| Useful | Info |
| :-- | :-- |
|Versions|Rancher `v2.4.3` UI: `v2.4.22` |
|Route|`undefined`|

giskou:
Fixes https://github.com/rancher/rancher/issues/27483

### DIFF
--- a/charts/rancher-monitoring/v0.1.1/charts/exporter-node-windows/templates/service.yaml
+++ b/charts/rancher-monitoring/v0.1.1/charts/exporter-node-windows/templates/service.yaml
@@ -11,6 +11,10 @@ metadata:
 spec:
   type: ClusterIP
   clusterIP: None
+  selector:
+    app: {{ template "app.name" . }}
+    chart: {{ template "app.version" . }}
+    release: {{ .Release.Name }}
   ports:
   - name: {{ .Values.ports.metrics.name }}
     port: {{ .Values.ports.metrics.port }}


PashCracken (dismissed):

recena:

PR #247, opened 2020-06-11
Release v2.4.5

deniseschannon:


charts/rancher-cis-benchmark/0.1.1/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.4.4-rc1
```
deniseschannon: Update this to 2.4.5-rc1. 

Update the max version for 0.1.0 to 2.4.4. We need to check if this will be an issue. 

@prachidamle

charts/rancher-monitoring/v0.1.1/questions.yml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.4.5-rc1
```
deniseschannon, line 1: We need to update the v0.1.0 chart to have a max version of v2.4.4

cbron (approved):

PR #254, opened 2020-06-17
[2.3]Disable monitoring v0.1.0

StrongMonkey:


charts/rancher-monitoring/v0.1.0/questions.yml
```diff
@@ -1 +1,2 @@
 rancher_min_version: 2.3.7-rc1
+rancher_max_version: 2.3.7-rc1
```
deniseschannon: This needs to be 2.3.7, otherwise it's not available for 2.3.7.

deniseschannon (changes requested):

charts/rancher-monitoring/v0.1.0/questions.yml
StrongMonkey: updated

deniseschannon (approved):

PR #264, opened 2020-06-22
add support for passing tolerations to log aggregators

paynejacob:
for: https://github.com/rancher/rancher/issues/20565

Added support for passing tolerations to logging aggregators.

### DIFF
--- a/charts/rancher-logging/0.2.2/.helmignore
+++ b/charts/rancher-logging/0.2.2/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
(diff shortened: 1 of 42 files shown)


charts/rancher-logging/0.2.0/charts/fluentd/charts/fluentd-linux/values.yaml
```diff
@@ -28,6 +28,8 @@ annotations: {}
 tolerations:
   - effect: NoSchedule
     operator: Exists
+  - effect: NoExecute
+    operator: Exists
```
cbron: Whats purpose of adding this ?

charts/rancher-logging/0.2.0/charts/fluentd/charts/fluentd-linux/values.yaml
```diff
@@ -41,12 +43,6 @@ service:
 cluster:
   dockerRoot: /var/lib/docker
 
-service:
-  ports:
-    - name: metric
-      port: 24231
-      targetPort: metric
-
```
cbron: Why removing metric service ?

cbron (changes requested):
* Does this need to happen in windows too ? 
* We never change charts in-place, we always do a copy of last chart, then commit, then any changes in 2nd commit.

charts/rancher-logging/0.2.0/charts/fluentd/charts/fluentd-linux/values.yaml
paynejacob: This was from when I was testing I will remove it

charts/rancher-logging/0.2.0/charts/fluentd/charts/fluentd-linux/values.yaml
paynejacob: again testing that should be removed

paynejacob:
This functionality exists in logging v2 on rancher 2.5.0. This version of logging is deprecated.

PR #276, opened 2020-07-01
Changes needed to use new security-scan image

prachidamle:
Adding a new version to the CIS chart to publish the new security-scan tag v0.1.12

### DIFF
--- a/charts/rancher-cis-benchmark/0.1.2/.helmignore
+++ b/charts/rancher-cis-benchmark/0.1.2/.helmignore
@@ -1,0 +1,22 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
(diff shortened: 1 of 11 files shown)


aiyengar2 (approved):
LGTM

leodotcloud:
We don't need a new chart version, just updating the image version in the existing one is enough.

leodotcloud (changes requested):
Please see previous comment.

prachidamle:
@leodotcloud we have already released the previous version in 2.4.5 release, hence changing the version 
cc @deniseschannon pls let me know if we can keep the version same if there is no upgrade impact

leodotcloud (approved):

leodotcloud:
Denise's comments:
we should always add a new system chart if any changes have existed
and increase the version
I understand that this chart is different, but it should follow the same paradigm of the other charts

dramich (approved):

PR #278, opened 2020-07-02
Istio 1.5.7

cbron:


charts/rancher-istio/1.5.700/charts/tracing/templates/service-jaeger.yaml
```diff
@@ -62,6 +62,26 @@ items:
     selector:
       app: jaeger
     type: ClusterIP
+- apiVersion: v1
+  kind: Service
+  metadata:
+    name: jaeger-collector-headless
```
cbron, line 68: This did not appear to exist in 1.4.x, so it was not removed in migration job.

charts/rancher-istio/1.5.700/templates/crd-certmanager-10.yaml
```diff
@@ -9,14 +9,14 @@ metadata:
     heritage: Tiller
     release: istio
   annotations:
-    helm.sh/resource-policy: keep
-    helm.sh/hook: crd-install
+    "helm.sh/resource-policy": keep
+    helm.sh/hook: crd-install,pre-upgrade
```
cbron, line 13: Note that I added pre-upgrade to all CRDs, previously they only used crd-install

charts/rancher-istio/1.5.700/templates/istio-onefive-migration.yaml
```diff
@@ -0,0 +1,77 @@
+---
```
cbron, line 2: This is the migration job

scripts/update-istio-release
```diff
@@ -4,37 +4,74 @@ set -e
...
+	grep -r --color=always --include "*.yaml" --exclude "*injection-template.yaml" "image: \"{{ .Values" charts/rancher-istio/$R_VERSION/
+	grep -r --color=always --include "*.yaml" "image: \"{{ \\$" charts/rancher-istio/$R_VERSION/
+	exit 0
+fi
+
+if [ -z "$(git status --porcelain)" ]; then
+  echo "making copy and committing"
+  cp -r charts/rancher-istio/$OLD_R_VERSION charts/rancher-istio/$R_VERSION
+  git add charts/rancher-istio/$R_VERSION
+  git commit -m "copying istio for new version $VERSION"
```
cbron, line 42: Note I am actually committing the copy version

charts/rancher-istio/1.5.700/templates/istio-onefive-migration.yaml
StrongMonkey: Will the migration job be always running no matter if it is fresh install or upgrade? Is it better to have an answer `migration: enabled` to control this?

StrongMonkey (commented):
The script is :fire: :+1:

charts/rancher-istio/1.5.700/templates/istio-onefive-migration.yaml
cbron: That's probably a good idea just to disable it in case its causing problems. But in reality it would alway run because we never know if 1.4 exists at upgrade time.

cbron: Added enabled flag

StrongMonkey: We can add it and it doesn't hurt. When this is no longer needed the flag can always be false. :+1:

charts/rancher-istio/1.5.700/templates/crd-certmanager-10.yaml
StrongMonkey: Is this for handling the case that Helm doesn't upgrade CRD?

cbron: That case, and also when a new CRD is added the upgrade doesn't install it because `crd-install` hook is not run on upgrade.

StrongMonkey: Got it.

brendarearden (dismissed):
🥇  here you go sir

StrongMonkey (approved):
:fire:

PR #279, opened 2020-07-03
remove grafana from istio chart to fix images.txt script

cbron:
When trying to generate rancher-images.txt I realized that we need to remove grafana image definition from sub-chart. Grafana defines their image the same way the script is looking for, and the script fails anything that matches that and isn't from rancher dockerhub.


brendarearden (approved):
LGTM

PR #283, opened 2020-07-09
Istio 1.5.8

cbron:
Removes 1.5.7, new security release usurps it.

https://github.com/istio/istio/compare/1.5.7...1.5.8

Interesting part is first commit


charts/rancher-istio/1.5.800/templates/istio-onefive-migration.yaml
```diff
@@ -59,7 +59,7 @@ metadata:
   annotations:
     "helm.sh/hook": pre-upgrade
     "helm.sh/hook-weight": "2"
-    "helm.sh/hook-delete-policy": hook-succeeded
```
StrongMonkey, line 62: What's the reason for changing this policy?

cbron: Only in order to debug easier. Removing on success cleans up resources, but this is a non-standard process so I figured it would be best to keep around for logs, shouldn't hurt to much to keep one job around. If you try to edit app it deletes as soon as upgrade starts.

brendarearden (approved):
LGTM

StrongMonkey (approved):

PR #285, opened 2020-07-11
[2.4] Release Istio 1.5.8

cbron:
These commits are cherry-picked and from a non dev-v2.4 branch because there are already other unrelated commits inside that dev-v2.4 branch.


deniseschannon:
Any way to clean up the commits @cbron?

cbron:
@deniseschannon squashed them all, re-tested locally just to double check and this works fine as long as we merge release-v2.4 back into dev-v2.4 immediately afterwards.

deniseschannon (approved):
Ideally we would have had 1 commit that was the copy from the previous version and a second commit with all the changes, but I didn't realize this was going into the release branch so it'll diverge from the dev branch. Ideally we should clean up dev-v2.4 to have the same type of commits for when we eventually bring dev-v2.4 into the release branch.

cbron:
@deniseschannon the dev branch does have those style of commits, it just had a few fixes on top of them and a change of version. Those are all already committed to the dev-2.4 branch so too late to clean up now without rewriting history and messing with peoples branches.

deniseschannon:
@cbron we could always revert them and then cherry pick in the new commits from this branch to clean up history and it won't mess anyone up. 

Knowing what the commit history is like. I'd say it might be cleaner to use the same commits you've already put into dev, otherwise trying to clean up dev for when we need to release 2.4.6 will be frustrating. 

Lesson learned is that next time, if we know we will be releasing something off cycle, we should be starting with a dedicated istio branch for testing and merge it into dev and release branches when it's being released.

cbron:
Merging now as full commits will appear in history from dev-v2.4 branch and this will sit on top.
Agree that this was messy way to handle and we need to branch next time instead of using dev-v2.4.

PR #287, opened 2020-07-12
[2.5] Forwardport Istio 1.5.8

cbron:
Squashed commits as best I could, but middle commits touched 1.5.7 so I can't remove it entirely.


StrongMonkey (approved):
LGTM

PR #291, opened 2020-07-21
Update eks-operator to v0.1.0-rc17

rmweir:


cbron (approved):
lgtm, shouldn't need chart bump since no real releases are happening.

PR #296, opened 2020-07-24
Make enabling/disabling Grafana reporting simpler

bentastic27:
Will at least make this a little more user friendly:

https://github.com/rancher/rancher/issues/28064

Didn't change the default as reporting_enabled still defaults to true.


recena:

PR #299, opened 2020-07-28
[v2.4] bump up rancher/fluentd version

aiwantaozi:
Problem:
configuration <match fluent.**> is deprecated after fluentd 1.x

Solution:
update default configuration

Issue:
https://github.com/rancher/rancher/issues/26543


cbron (approved):
lgtm, this is only logging version going into 2.4.6

PR #303, opened 2020-07-29
bump up rancher/fluentd version to v0.1.21

aiwantaozi:
Problem:
We already add label in rancher FLUENT_LOG, if we also add FLUENT_LOG here will get duplicated configuration error.

Solution:
Remove FLUENT_LOG label.

Issue:
https://github.com/rancher/rancher/issues/26543


aiwantaozi:
@cbron Could you help with it, thank you?

cbron:
Closing in favor of removing chart since no other changes: https://github.com/rancher/system-charts/pull/304

PR #308, opened 2020-07-31
Add 2.4.99 as max version for OPA as it's moved to charts repo

deniseschannon:


cbron (dismissed):
lgtm, we'll have to remember this in code-review for future charts, but if people just copy previous version then it will stay there.

cbron (approved):
Lgtm, will merge later if Denise doesn’t sooner

PR #309, opened 2020-08-01
Add eks-operator

rmweir:


charts/rancher-eks-operator/0.1.0/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.0-rc1
```
deniseschannon: I wonder if this will cause issues in the alpha and maybe this should be `2.5.0-alpha1`

deniseschannon (commented):
Besides the min version, I want this moved into dev-v2.4 to consolidate branches.

charts/rancher-eks-operator/0.1.0/questions.yaml
rmweir: I think maybe it sees the "-" in a version and just treats it as a prelease, where its equal to any other prerelease but just in case I can change to alpha

deniseschannon (approved):

cbron (approved):

PR #317, opened 2020-08-11
RKE2

briandowns:
Adds necessary components to be able to run security-scan for RKE2.

If approved, I'll squash merge.


brandond (approved):

prachidamle (approved):

cbron:
This change needs to be reverted and instead added to a new chart version, once a version has gone live in the [release branch](https://github.com/rancher/system-charts/tree/release-v2.4/charts/rancher-cis-benchmark/0.1.2) we don't ever edit it or users who have installed that chart would never get the change, and we wouldn't know which customers have which versions installed. So we add 1 new chart version per rancher release as necessary. 
For adding system-chart versions, the first commit should always be a copy of the existing chart, and the 2nd commit should be the changes involved, that way we can review changes easily. Ex: https://github.com/rancher/system-charts/pull/265/commits

PR #320, opened 2020-08-14
Restore namespace of metrics when available

scoof:
This will enable federation of these statistics to the project monitoring Prometheus.

See https://github.com/rancher/rancher/issues/28394


scoof:
Closing, since this is not applicable for the new prom operator setup

PR #321, opened 2020-08-24
bump K3s upgrader version to 0.2.1

MonzElmasry:
When helm uninstall command is executed, cattle-system namespace gets deleted which causes the cattle agent to disconnect, so by removing the namespace.yaml from this chart only K3s-upgrader app is deleted without deleting cattle-system ns

### DIFF
--- a/charts/rancher-k3s-upgrader/0.2.1/Chart.yaml
+++ b/charts/rancher-k3s-upgrader/0.2.1/Chart.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+name: rancher-k3s-upgrader
+description: Enables a k3s or rke2 cluster to update itself by reacting to Plan CRs.
+  Users do not need to manually upgrade this app. It will be automatically upgraded to the latest version when upgrading a cluster.
+home: https://github.com/rancher/system-charts/blob/dev-v2.4/charts/rancher-k3s-upgrader
+sources:
+  - "https://github.com/rancher/system-charts/blob/dev-v2.4/charts/rancher-k3s-upgrader"
+version: 0.2.0
+appVersion: v0.6.2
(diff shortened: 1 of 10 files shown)


cbron (approved):
lgtm

dweomer (approved):

cjellick:
I ran this PR past @ibuildthecloud because it makes me nervous.

We both don't understand WHY this doesn't delete the namespace and I kind of want to understand that a bit. I'm not even sure to tell you where to look to figure that out though.

Maybe @brendarearden knows

brendarearden:
Is there a reason we didn't notice this happening with the first version of this chart or is this first time we have tested uninstall for this chart?

Basically if you remove the namespace template from your chart, then it doesn't recognize the namespace as a resource it created, so it won't remove it. (info on [uninstall](https://helm.sh/docs/helm/helm_uninstall/))

cjellick:
> Is there a reason we didn't notice this happening with the first version of this chart or is this first time we have tested uninstall for this chart?

bc this chart never gets explicitly uninstalled. its a system chart that is installed and then basically lives for the life of the cluster. to repro the bug, user has to explicitly go in and delete the app from the system project.

> Basically if you remove the namespace template from your chart, then it doesn't recognize the namespace as a resource it created, so it won't remove it. (info on uninstall)

A couple thoughts here:
1. this seeks to correct the issue by causing the app to upgrade to this version. the question at hand is why doesnt it delete the NS as part of the *upgrade*. it apparently doesnt, but not understanding why is the concerning part
2. Regarding uninstall, I thought helm calculates what to delete based on the snapshot of the last deployed release that it stored in a secret, not based on what is in the chart itself.

brendarearden:
> 1. this seeks to correct the issue by causing the app to upgrade to this version. the question at hand is why doesnt it delete the NS as part of the _upgrade_. it apparently doesnt, but not understanding why is the concerning part

Okay I see what you are saying. It would delete on upgrade. Maybe we should change the approach and use `metadata:
  annotations:
    "helm.sh/resource-policy": keep` on the namespace so it won't be deleted. 

> 2. Regarding uninstall, I thought helm calculates what to delete based on the snapshot of the last deployed release that it stored in a secret, not based on what is in the chart itself.

This is correct, you do not have to supply the chart for uninstall. I was thinking of this situation on a fresh install, not on an upgrade then delete, so that was my bad.

cjellick:
In all of @MonzElmasry's testing, it DOES NOT delete the NS on upgrade. I think it is because of the helm 2 behavior described here:
https://helm.sh/docs/faq/#automatically-creating-namespaces

i think helm2 is figuring out that it shouldnt delete the NS on upgrade because it is referenced in the chart.

cjellick:
@MonzElmasry you need to link the issue in this pr

PR #327, opened 2020-08-28
Update istio tracing chart to use a serviceaccount as other istio components

dsim62:
As we use pod security policies and we don't want to use default serviceaccount of namespace istio-system for security reasons, please use serviceaccount for istio tracing  as other istio components

Thank you,

Regards,

David


sgall62 (approved):

recena:

PR #338, opened 2020-09-17
Do not uninstall CRD

rmweir:
**Problem:**
Deleting the eks-operator app deletes the eksclusterconfig crd which causes eks cluster that are up at deletion to start deleting and prevents new eks clusters from being created.

**Solution:**
Do not delete crd on app deletion. Crd will still be updated on upgrade.

**Issue:**
https://github.com/rancher/rancher/issues/28535

### DIFF
--- a/charts/rancher-eks-operator/0.1.0/templates/eksclusterconfig.yaml
+++ b/charts/rancher-eks-operator/0.1.0/templates/eksclusterconfig.yaml
@@ -2,6 +2,8 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: eksclusterconfigs.eks.cattle.io
+  annotations:
+    "helm.sh/resource-policy": keep
 spec:
   group: eks.cattle.io
   names:


paynejacob (commented):
What happens if a user is trying to uninstall the chart? won't this leave the crd?

brendarearden (approved):
that should do the trick! 

Note: it looks like there was an issue in helm 3 with versions prior to helm 3.2.0 with upgrade not respecting resource policies. Just keep that in mind for if we do move this chart to helm 3.

cbron (commented):
Yes uninstalling the chart will leave the CRDs/CRs. Thats ok, as long as they can re-install the chart and the two things don't conflict, have you tested that ricky ? I guess QA will. I don't see how this would stop it, unless helm says it can't take them over because it didn't install them. Same situation if the new chart CRD is different than original CRD still installed, hoping that won't conflict.

rmweir:
@cbron I tested deleting the app then reinstalling and upgrading the app. Both while have a cluster in the middle of provisioning. Both were successful.

cbron (approved):

paynejacob (approved):

PR #346, opened 2020-09-30
add max version to k3s upgrader

cbron:


cbron:
closing for https://github.com/rancher/system-charts/pull/347

PR #355, opened 2020-10-09
Add env vars to eks-operator for proxy support

rmweir:
**Problem:**
Eks-operator is not working in airgap proxy setups.

**Solution:**
Add fields for configuring proxy environment variables in deployment.

**Issue:**
https://github.com/rancher/rancher/issues/29070

### DIFF
--- a/charts/rancher-eks-operator/1.0.101/templates/NOTES.txt
+++ b/charts/rancher-eks-operator/1.0.101/templates/NOTES.txt
@@ -1,0 +1,4 @@
+You have deployed the Rancher EKS operator
+Version: {{ .Chart.AppVersion }}
+Description: This operator provisions EKS clusters
+from EKSClusterConfig CRs.
--- a/charts/rancher-eks-operator/1.0.101/templates/deployment.yaml
+++ b/charts/rancher-eks-operator/1.0.101/templates/deployment.yaml
@@ -1,0 +1,20 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: eks-config-operator
+  namespace: cattle-system
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      ke.cattle.io/operator: eks
+  template:
+    metadata:
+      labels:
+        ke.cattle.io/operator: eks
+    spec:
+      serviceAccountName: eks-operator
+      containers:
+      - name: eks-operator
+        image: {{ template "system_default_registry" . }}{{ .Values.eksOperator.image.repository }}:{{ .Values.eksOperator.image.tag }}
+        imagePullPolicy: IfNotPresent
--- a/charts/rancher-eks-operator/1.0.101/templates/deployment.yaml
+++ b/charts/rancher-eks-operator/1.0.101/templates/deployment.yaml
@@ -18,3 +18,10 @@ spec:
       - name: eks-operator
         image: {{ template "system_default_registry" . }}{{ .Values.eksOperator.image.repository }}:{{ .Values.eksOperator.image.tag }}
         imagePullPolicy: IfNotPresent
+        env:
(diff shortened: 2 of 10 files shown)


rmweir:
The only difference is in the deployment yaml. HTTPS_PROXY, HTTP_PROXY, NO_PROXY are now passed as env variables and are configurable through values.

charts/rancher-eks-operator/1.0.101/templates/NOTES.txt
```diff
@@ -0,0 +1,4 @@
+You have deployed the Rancher EKS operator
+Version: {{ .Chart.AppVersion }}
+Description: This operator provisions EKS clusters
+from EKSClusterConfig CRs.
```
nickgerace, line 4: Perhaps there's a rule I don't know; is this line wrapping intentional?

charts/rancher-eks-operator/1.0.101/templates/deployment.yaml
```diff
@@ -0,0 +1,27 @@
...
+  template:
+    metadata:
+      labels:
+        ke.cattle.io/operator: eks
+    spec:
+      serviceAccountName: eks-operator
+      containers:
+      - name: eks-operator
+        image: {{ template "system_default_registry" . }}{{ .Values.eksOperator.image.repository }}:{{ .Values.eksOperator.image.tag }}
+        imagePullPolicy: IfNotPresent
```
nickgerace, line 20: Would there be an advantage to having the pull policy in ```values.yaml``` instead?

nickgerace (dismissed):

charts/rancher-eks-operator/1.0.101/templates/deployment.yaml
rmweir: Probably not, we have no need to configure it at the moment.

charts/rancher-eks-operator/1.0.101/templates/NOTES.txt
rmweir: Probably to be compatible with default printer column widths for commands such as `helm search repo`: https://helm.sh/docs/helm/helm_search_repo/#options

mrajashree (dismissed):

cbron:
For adding system-chart versions, the first commit should always be a copy of the existing chart, and the 2nd commit should be the changes involved, that way we can review changes easily. Ex: https://github.com/rancher/system-charts/pull/265/commits

rmweir:
@cbron oh that definitely simplifies it. Updated.

charts/rancher-eks-operator/1.0.101/templates/deployment.yaml
```diff
@@ -18,3 +18,10 @@ spec:
       - name: eks-operator
         image: {{ template "system_default_registry" . }}{{ .Values.eksOperator.image.repository }}:{{ .Values.eksOperator.image.tag }}
         imagePullPolicy: IfNotPresent
+        env:
+        - name: HTTP_PROXY
+          value: {{ .Values.httpProxy }}
```
cbron, line 23: Nit: Seems that it is ok if these are set to blank, but you could wrap in conditional so they aren't set at all if blank.

cbron (approved):
Questions min/max should be fine as-is.

charts/rancher-eks-operator/1.0.101/templates/deployment.yaml
rmweir: The rancher code marshal the struct into yaml and passes it to helm without checking if each variable is empty. Therefore, they will never be empty as rancher uses it. I can do the check though.

cbron: Na its fine

PR #362, opened 2020-10-20
[2.4] updated fluentd containers to use the latest rancher image

paynejacob:


nickgerace:
Slightly unrelated, do we need backports for the tolerations and nodeselector changes to this chart?

nickgerace (dismissed):

brendarearden:
When adding system charts, there should be at least two separate commits. The first commit should be just a base copy of the existing chart, and then the second commit would be the changes for the new version. It makes it easier to quickly see the changes from the previous version. Ex: https://github.com/rancher/system-charts/pull/265/commits - please update your pr to follow this flow

nickgerace (approved):

brendarearden (approved):
lgtm

paynejacob:
This is waiting on a fluentd release

paynejacob:
This is stale and not needed

PR #363, opened 2020-10-23
add cni mount for use by security-scan

briandowns:
Added additional volume mount so security-scan has access to RKE2 CNI.

Signed-off-by: Brian Downs <brian.downs@gmail.com>

### DIFF
--- a/charts/rancher-cis-benchmark/0.1.3/templates/pod.yaml
+++ b/charts/rancher-cis-benchmark/0.1.3/templates/pod.yaml
@@ -45,6 +45,10 @@ spec:
       hostPath:
         path: /var/lib/rancher
         type: Directory
+    - name: rke2-cni
+      hostPath:
+        path: /etc/cni/net.d
+        type: Directory
   containers:
     - name: {{ .Chart.Name }}
       restartPolicy: Never
@@ -97,6 +101,8 @@ spec:
         {{- end }}
         - mountPath: /var/lib/rancher
           name: rke2
+        - mountPath:
+          name: rke2-cni
       resources:
         {{- toYaml .Values.resources | nindent 12 }}
   {{- with .Values.nodeSelector }}


cbron:
What release is this going into, is there a linked issue ? 2.4.9 has already gone through QA testing so I'm not sure we want to merge now.

brandond (approved):

cjellick:
Should not go into 2.4; just needs supported in 2.5. Where should such changes for "cis scan v2" go? I think they are ~still driven through system-charts as there is no scan chart in rancher/charts~. @prachidamle 
_edit: there is a chart for it in rancher/charts_

@cbron is there a confluence article or something that lays out the branching and release paradigm for rancher/system-charts and rancher/charts? I've notice that it is quite difficult for non-core rancher2 folks to grok where/how to make changes.

cjellick:
So, is this PR not needed if we are only supporting RKE2 scans in 2.5?

briandowns:
> So, is this PR not needed if we are only supporting RKE2 scans in 2.5?

If only 2.5 is needed, then no. I would prefer to keep the branch around though for reference but also in case we decide that it'd be needed in 2.4.

prachidamle:
yes, @cjellick @briandowns (as we discussed yesterday on zoom) we will not need this PR if we are not supporting Rancher v2.4
If we we want to support rke2 only in v2.5 - we don't need changes done to rancher/system-charts - these are only for release/v2.4.

cc @cbron

prachidamle:
@briandowns you will have to put these changes to the cis-operator for v2.5: 
Add the volumes to the PodSpec here https://github.com/rancher/cis-operator/blob/master/pkg/securityscan/job/job.go#L78

cbron:
@briandowns  / @prachidamle should we undo the 0.1.3 version of CIS benchmark then ? https://github.com/rancher/system-charts/commits/dev-v2.4/charts/rancher-cis-benchmark/0.1.3

PR #365, opened 2020-10-28
release-v2.4

kinarashah:


luthermonson (approved):

charts/rancher-monitoring/v0.1.4/questions.yml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.4.9-rc1
```
cbron, line 1: We _do_ apparently want this to be available in 2.4.9, previously only 0.1.2 was available there.

charts/rancher-eks-operator/1.0.101/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+name: rancher-eks-operator
+description: A Helm chart for provisioning EKS clusters
+home: https://github.com/rancher/eks-operator
+sources:
+  - "https://github.com/rancher/eks-operator"
+version: 1.0.101
+appVersion: v1.0.1
```
cbron, line 8: We might want to remove eks-operator from 2.4 line eventually since we made a 2.5 line, but its fine for now since questions.yaml is in place.  cc @rmweir @thedadams

cbron (approved):

PR #336, opened 2020-09-16
Update rancher-external-dns images

pennyscissors:
**This PR includes:**
- Update external-dns image to v0.7.3
- Add endpoint resource to clusterrole since they are now required
- Bump chart version to v0.1.1

**Related PRs:**
https://github.com/rancher/image-mirror/pull/35

PR #370, opened 2020-11-09
Add batch size to avoid rate limits

dkeightley:
Avoid [Route53 API rate limits](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html#limits-api-requests) when large changes occur in a cluster, using a batch size of `100` as per https://github.com/zalando-incubator/kubernetes-on-aws/pull/2266/files

Related: https://github.com/rancher/rancher/issues/28391

### DIFF
--- a/charts/rancher-external-dns/v0.1.1/templates/deployment.yaml
+++ b/charts/rancher-external-dns/v0.1.1/templates/deployment.yaml
@@ -80,6 +80,9 @@ spec:
           {{- if .Values.google.project }}
             - --google-project={{ .Values.google.project }}
           {{- end }}
+          {{- if eq .Values.provider "aws" }}
+            - --aws-batch-change-size=100
+          {{- end }}
           {{- if eq .Values.provider "infoblox" }}
             - --infoblox-grid-host={{ .Values.infoblox.gridHost }}
             {{- if .Values.infoblox.domainFilter }}
--- a/charts/rancher-external-dns/v0.1.1/templates/deployment.yaml
+++ b/charts/rancher-external-dns/v0.1.1/templates/deployment.yaml
@@ -81,7 +81,7 @@ spec:
             - --google-project={{ .Values.google.project }}
           {{- end }}
           {{- if eq .Values.provider "aws" }}
-            - --aws-batch-change-size=100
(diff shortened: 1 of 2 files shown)


charts/rancher-external-dns/v0.1.1/templates/deployment.yaml
```diff
@@ -80,6 +80,9 @@ spec:
           {{- if .Values.google.project }}
             - --google-project={{ .Values.google.project }}
           {{- end }}
+          {{- if eq .Values.provider "aws" }}
+            - --aws-batch-change-size=100
+          {{- end }}
```
prachidamle, line 85: Instead of setting this value to 100, we should make it configurable by exposing in Values.yaml and set to 100 in Values

```
          {{- if eq .Values.provider "aws" }}
            - --aws-batch-change-size={{ .Values.aws.batchChangeSize }}
          {{- end }}
```

prachidamle:
Note that this change will not expose the parameter in Rancher UI though. It will just make sure its set to a lower value when the chart is deployed. To expose in Rancher UI, we will need to change Rancher API https://github.com/rancher/types/blob/e1dd38de6b4a7dd4808b8de4d6d569ff71e2d8e2/apis/management.cattle.io/v3/globaldns_types.go#L54

charts/rancher-external-dns/v0.1.1/templates/deployment.yaml
dkeightley: Agree, thanks for the suggestion

prachidamle (approved):
LGTM

dkeightley:
I'd be happy to backport this for the v2.4.x branch and v2.3.x (chart v0.0.2) if desired.

cbron (commented):
This change seems fine/correct, but I don't like the idea of only adding it to the chart and not the UI. I'll see how hard its going to be to add it. 

In the meantime, @dkeightley can you first do a commit that is only a copy of the last chart, and then make all changes in the new version along with version bumps. It should look like this: https://github.com/rancher/system-charts/pull/336/commits

And then once both of the above are figured out we can do a PR to 2.4

cbron:
Closing in favor of other PRs

PR #374, opened 2020-11-20
Increasing replicas to allow nodes to drain with istio enabled

### LINKED ISSUE #30082: Nodes can not be drained if Istio is enabled
Due to the disruption budget, we can not drain nodes in a cluster with Istio deployed. We should update the istio overlay to have a min of 2 replicas of pilot and gateways.  Hypothetically this should allow nodes to be successfully drained.

**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Deploy cluster
2. Enable Istio with ingress gateways
3. Drain a node

**Result:**
The node will timeout and remain cordoned but not drained.

**Other details that may be helpful:**
Updating the istio overlay to have a min of 2 replicas of pilot and gateways may help work around this.
I've tried setting `galley.replicaCount` to `2` and `mixer.policy.replicaCount` to `2` in the app's answers then upgrading. Though they do not appear in the values.yml in the chart, the configurations appear to have been accepted (but nodes still do not drain).  Scaling pilot to 2 manually hung.

It's been proposed that we may be able to fix this by adding an overlay file with the following contents:

```
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  components:
    ingressGateways:
    - enabled: true
      k8s:
        hpaSpec:
          minReplicas: 2
    pilot:
      enabled: true
      k8s:
        hpaSpec:
          minReplicas: 2
```

See: https://github.com/istio/istio/issues/12602#issuecomment-701038118

I'm not sure yet how to test this.

**Environment information**
- Rancher version: v2.4.5
- Installation option (single install/HA): single docker
- Kubernetes version (use `kubectl version`): 1.18

Tejeev:
mixer.telemetry.autoscaleMin=2
mixer.telemetry.replicaCount=2
mixer.policy.autoscaleMin=2
mixer.policy.replicaCount=2
pilot.autoscaleMin=2
galley.replicaCount=2
security.replicaCount=2
sidecarInjectorWebhook.replicaCount=2

fixes  rancher/rancher#30082

### DIFF
--- a/charts/rancher-istio/1.5.900/charts/galley/values.yaml
+++ b/charts/rancher-istio/1.5.900/charts/galley/values.yaml
@@ -2,7 +2,7 @@
 # galley configuration
 #
 enabled: true
-replicaCount: 1
+replicaCount: 2
 rollingMaxSurge: 100%
 rollingMaxUnavailable: 25%
 image: galley
--- a/charts/rancher-istio/1.5.900/charts/mixer/values.yaml
+++ b/charts/rancher-istio/1.5.900/charts/mixer/values.yaml
@@ -10,9 +10,9 @@ env:
 policy:
   # if policy is enabled, global.disablePolicyChecks has affect.
   enabled: false
-  replicaCount: 1
+  replicaCount: 2
   autoscaleEnabled: true
-  autoscaleMin: 1
+  autoscaleMin: 2
   autoscaleMax: 5
   cpu:
     targetAverageUtilization: 80
@@ -21,9 +21,9 @@ policy:
 
 telemetry:
   enabled: true
-  replicaCount: 1
+  replicaCount: 2
   autoscaleEnabled: true
-  autoscaleMin: 1
+  autoscaleMin: 2
   autoscaleMax: 5
   cpu:
     targetAverageUtilization: 80
--- a/charts/rancher-istio/1.5.900/charts/pilot/values.yaml
+++ b/charts/rancher-istio/1.5.900/charts/pilot/values.yaml
@@ -3,7 +3,7 @@
 #
 enabled: true
 autoscaleEnabled: true
-autoscaleMin: 1
+autoscaleMin: 2
 autoscaleMax: 5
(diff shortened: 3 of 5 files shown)


charts/rancher-istio/1.5.900/charts/galley/values.yaml
```diff
@@ -2,7 +2,7 @@
 # galley configuration
 #
 enabled: true
-replicaCount: 1
+replicaCount: 2
```
cbron, line 5: This is really caused by this: https://github.com/istio/istio/issues/12602#issuecomment-489449657
I don't think just bumping the replica count for everyone is the right solution though, even though its a good workaround. Istio is already heavy as it is, this could crash existing clusters. Perhaps we could lower/delete the disruption budget ? On the other side, this version is deprecated and there is a workaround, so we shouldn't put too much time into fixing a problem that the upstream team won't fix.

brendarearden (approved):
LGTM

charts/rancher-istio/1.5.900/charts/galley/values.yaml
Tejeev: That makes sense to me.  Could we maybe add a configurable for the disruption budget that manages the minimum scale accordingly for the user and lets them know of the increased resource requirements?

Tejeev: I think it's worth having a fix for this as it's... disruptive.. (sorry, not sorry)

cbron: Would a valid fix be to turn `defaultPodDisruptionBudget.enabled` to false ? Changing the PDB `minAvailable` itself is discouraged by istio because people would try to change it on update and its immutable. 
Beyond that, if you mean a single variable to update the scale of each of these pods themselves, the chart doesn't do that right now. We could add it but it feels kinda weird.

recena:

PR #392, opened 2021-01-04
Update istio injection template for limits

paynejacob:
https://github.com/rancher/rancher/issues/28950

PROBLEM:
Rancher's Istio fork was missing upstream changes that set resource limits on the sidecar pods.

SOLUTION:
Copy the change from upstream to read the annotations and apply the limits to the sidecar.

TESTING:
See issue https://github.com/rancher/rancher/issues/28950


charts/rancher-istio/1.5.1000/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+appVersion: 1.5.9
```
superseb: Do `appVersio`/`version` need to be adjusted for the new version?

brendarearden: are we adding support for istio 1.5.10 or are we making changes to istio 1.5.9? If we are making changes to 1.5.9, the Version on line 17 needs updated and we should be using 1.5.901, not 1.5.1000.

paynejacob: We are changing 1.5.9 ill fix the versions

brendarearden (changes requested):
Update version and overall file version to 1.5.901 since we are not adding support for 1.5.10, and just making changes to 1.5.9

charts/rancher-istio/1.5.901/Chart.yaml
```diff
@@ -0,0 +1,17 @@
...
+- security
+- sidecarInjectorWebhook
+- mixer
+- pilot
+- galley
+name: rancher-istio
+sources:
+- http://github.com/istio/istio
+tillerVersion: '>=2.7.2-0'
+version: 1.5.900
```
brendarearden: version should also be increased from `1.5.900` to `1.5.901`

brendarearden (changes requested):

superseb (dismissed):

cbron (changes requested):
The issue is for 2.4, but this is going into dev-v2.5. 
Does this code exist in upstream 1.5.9 codebase ?

brendarearden:
@cbron They do exist in the 1.5.9 https://github.com/istio/istio/blob/1.5.9/manifests/istio-control/istio-autoinject/files/injection-template.yaml#L353 

the original issue makes it sound like these exact limits existed since 1.4.x, but that is not the case. We have had the limits that were merged for in istio 1.4.x in our rancher charts since they were merged upstream. 

We did not update our 1.5 chart when the additional limits were added in (starting in 1.5.2) and this file was recently updated as well with https://github.com/istio/istio/pull/29729/files

charts/rancher-istio/1.5.901/files/injection-template.yaml
```diff
@@ -348,15 +348,24 @@ containers:
...
-  resources:
-    {{ if or (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyCPU`) (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyMemory`) -}}
-    requests:
-      {{ if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyCPU`) -}}
-      cpu: "{{ index .ObjectMeta.Annotations `sidecar.istio.io/proxyCPU` }}"
-      {{ end}}
-      {{ if (isset .ObjectMeta.Annotations `sidecar.istio.io/proxyMemory`) -}}
-      memory: "{{ index .ObjectMeta.Annotations `sidecar.istio.io/proxyMemory` }}"
-      {{ end }}
+resources:
```
brendarearden, line 351: formatting looks off - this will cause an issue , unless you confirmed and its just a github ui issue.

PR #393, opened 2021-01-06
[2.5.4] Release

paynejacob:


cbron:
Did we confirm only one new version of each going in ? Looks like it. Not sure if any old ones need to be deprecated, sounds like this monitoring version is right since 0.1.4 still needs to be available.

aiyengar2 (commented):
#### Note

For rancher-monitoring, v0.2.0 and v0.1.4 are both being released with no max version since v0.2.0 cannot be installed on a k8s 1.15 cluster.

v0.2.0 has a rancher-min-version of 2.5.4 (when Rancher started to honor kubeVersions on charts) and v0.1.4 remains the same as v0.1.3 (since kubeVersion is not used). 

A user on a k8s 1.16+ cluster should be able to install both, although they are advised to use v0.2.0.

cbron (approved):
Should be ok when we are ready to release 2.5.4, but wait until then.

PR #390, opened 2020-12-21
[2.4] Update logging to 0.2.3 for Windows 1909 support

nickgerace:
Original PR: https://github.com/rancher/system-charts/pull/384
Relevant QA issue: https://github.com/rancher/rancher/issues/30569
Relevant issue: rancher/rancher#30279

PR #395, opened 2021-01-11
Upgrade metadata filter

rmweir:
Update logging to use image that includes fluentd metadata filter upgrade: https://github.com/rancher/fluentd/pull/37

**Issue:**
https://github.com/rancher/rancher/issues/30425

### DIFF
--- a/charts/rancher-logging/0.2.4/charts/fluentd/charts/fluentd-linux/values.yaml
+++ b/charts/rancher-logging/0.2.4/charts/fluentd/charts/fluentd-linux/values.yaml
@@ -1,0 +1,56 @@
+nameOverride: fluentd
+labels: {}
+
+image:
+  repository: rancher/fluentd
+  tag: v0.1.22
+  pullPolicy: IfNotPresent
+
+resources: {}
+  # limits:
+  #   cpu: 100m
+  #   memory: 500Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 200Mi
+# env:
+
+command: '["fluentd", "-c", "/fluentd/etc/config/entry/fluent.conf"]'
+
+livenessProbe:
+  enabled: true
+
+annotations: {}
+
+# updateStrategy:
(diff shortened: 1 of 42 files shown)


nickgerace (commented):
@rmweir did we not want to increment the Windows image tag as well? The manifest should be functionally identical, but I can also see why we would not want to increment.

rmweir:
@nickgerace I will go ahead and increment although this only modifies the linux gems so it will have no effect on windows I imagine.

nickgerace (dismissed):

charts/rancher-logging/0.2.4/charts/fluentd/charts/fluentd-linux/values.yaml
```diff
@@ -54,3 +54,4 @@ configmapReload:
     tag: v0.3.0-rancher3
     pullPolicy: IfNotPresent
   resources: {}
+
```
cbron, line 57: What about fluentd-tester ? See https://github.com/rancher/system-charts/pull/390/commits/4d4e3c2d419920f9559c1d1a90374e18835dd015#diff-7fbe58cc263dd407812a62df27b04a8a2b700eb11e30a24f8d6d1744c2e43dcfR5

cbron: @nickgerace I assume we need to bump this so tester and real fluentD don't act differently, even though this specific bug probably won't show up there

nickgerace: Ah tester would actually be necessary since it is specifically for fluentd Linux, if I recall correctly. @cbron

rmweir: Updated

cbron (approved):

PR #397, opened 2021-01-13
Istio Sidecar Limits

paynejacob:
rancher/rancher#28950

PROBLEM:
Rancher's Istio fork was missing upstream changes that set resource limits on the sidecar pods.

SOLUTION:
Copy the change from upstream to read the annotations and apply the limits to the sidecar.

TESTING:
See issue rancher/rancher#28950


brendarearden (approved):
LGTM

superseb (approved):

PR #405, opened 2021-01-28
[2.5.7] Introducing aks-operator Helm chart

mjura:


charts/rancher-aks-operator/1.0.100/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.0-alpha1
```
rmweir: I believe this should be 2.5.7-alpha1.

charts/rancher-aks-operator/1.0.100/templates/clusterrole.yaml
```diff
@@ -0,0 +1,15 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: aks-operator
+  namespace: cattle-system
+rules:
+  - apiGroups: ['']
+    resources: ['secrets']
+    verbs: ['get', 'list', 'create', 'watch']
```
rmweir, line 9: Pending changes to aks-operator we may be able to remove some or all of these secrets permissions in the future.

rmweir (approved):
LGTM. Keep track of permissions used in aks-operator so we can remove anything not needed. Left a comment regarding this. Will need to replace image tag once there is actually an aks-operator tag available. Can probably wait to merge until that is done.

cbron (commented):
lgtm, waiting for 2.5.6 release

cbron (approved):

PR #407, opened 2021-02-01
Set rancher_min_version to 2.5.6 for eks-operator 1.0.6

thedadams:
Issue:
https://github.com/rancher/rancher/issues/31062


rmweir (approved):
LGTM

cbron (approved):
I think this PR is fine, the real dev-side question here is can we confirm that the eks-operator won't auto-deploy with this in place. So we need to ensure we QA that.

aiyengar2 (approved):
LGTM

PR #409, opened 2021-02-03
Update wmi_exporter to v0.0.5

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/27911

Why is it related?

In Feb 11, 2020 Windows rolled out a security update that changed the
interface between user mode and kernel mode, so difficult-to-debug issues
related to starting the container process can show up if the Windows Server
container image and the Windows container host are out of sync with the
patch.

This is most likely the case with the AWS AMI that experienced the issue
noted on the ticket, so the fix from our end is to bump the version of
wmi_exporter to one that was built after the patch was released.

Link to Announcement:
https://support.microsoft.com/en-us/topic/you-might-encounter-issues-when-using-windows-server-containers-with-the-february-11-2020-security-update-release-b9a8fcae-950d-7a0b-ac7c-cb6b294cb809


charts/rancher-monitoring/v0.1.5/values.yaml
```diff
@@ -172,7 +172,7 @@ exporter-node-windows:
   image:
     os: windows
     repository: rancher/wmi_exporter-package
-    tag: v0.0.3
+    tag: v0.0.4
```
aiyengar2: @luthermonson I can bump this to v0.0.5 but the manifest was never created (e.g. https://hub.docker.com/v2/repositories/rancher/wmi_exporter-package/tags/v0.0.5 does not exist)

Seems like this was because the 20H2 build failed (https://drone-publish.rancher.io/rancher/wmi_exporter-package/13/5/2); could you look into it?

If it's not a difficult fix, I'd ideally like to bump to the latest before pushing out 2 new charts.

aiyengar2 (commented):
#### Note to Reviewers

This PR has 2 charts since Monitoring V1 maintains two separate lines; one to support k8s 1.15 cluster users in Rancher 2.4 and another to support k8s 1.16+ users.

charts/rancher-monitoring/v0.1.5/values.yaml
nickgerace: I'll approve, but want to see if `0.0.5` is possible.

nickgerace (dismissed):

aiyengar2:
@nickgerace bumped PR to v0.0.5, waiting on https://github.com/rancher/wmi_exporter-package/pull/8 for drone fix and for the image to be available at https://hub.docker.com/v2/repositories/rancher/wmi_exporter-package/tags/v0.0.5.

charts/rancher-monitoring/v0.1.5/values.yaml
aiyengar2: bumped

nickgerace (approved):

aiyengar2:
v0.0.5 manifest now exists and a 20H2 image has been pushed out.

luthermonson (approved):

dramich (approved):

PR #413, opened 2021-02-12
Update default kubelet path for k8s 1.20

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/31167, https://github.com/rancher/rancher/issues/31305

### DIFF
--- a/charts/rancher-monitoring/v0.2.1/charts/exporter-kubelets/templates/_helpers.tpl
+++ b/charts/rancher-monitoring/v0.2.1/charts/exporter-kubelets/templates/_helpers.tpl
@@ -1,8 +1,10 @@
 {{- define "kubelet.serviceMonitor.resourcePath" -}}
 {{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
-{{- if semverCompare ">=1.20.0-0" $kubeTargetVersion -}}
-{{ default .Values.kubelet.serviceMonitor.resourcePath "/metrics/resource" }}
+{{- if .Values.kubelet.serviceMonitor.resourcePath -}}
+{{ .Values.kubelet.serviceMonitor.resourcePath }}
+{{- else if semverCompare ">=1.20.0-0" $kubeTargetVersion -}}
+/metrics/resource
 {{- else -}}
-{{ default .Values.kubelet.serviceMonitor.resourcePath "/metrics/resource/v1alpha1" }}
+/metrics/resource/v1alpha1
 {{- end -}}
 {{- end }}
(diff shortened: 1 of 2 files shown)


aiyengar2 (commented):
v0.2.1 has not been released yet in release-v2.5, so no version bump is necessary.

This also effectively backports the change introduced in https://github.com/rancher/charts/pull/992.

pennyscissors (dismissed):

charts/rancher-monitoring/v0.2.1/charts/exporter-kubelets/templates/_helpers.tpl
```diff
@@ -0,0 +1,8 @@
+{{- define "kubelet.serviceMonitor.resourcePath" -}}
+{{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
+{{- if semverCompare ">=1.20.0-0" $kubeTargetVersion -}}
```
cbron: I'm not sure how well this is actually working, there have been quite a few complaints in that original issue around limiting deployment to 1.20. Do we not want to use APICapabilities because both would get returned ? I have seen both ways done before, it would be good to have a pattern.

aiyengar2: We could use .Capabilities.APIVersions, but this is the way that kube-prometheus-stack has done it for dashboards before:

https://github.com/prometheus-community/helm-charts/blob/29df8b9de95c258d35d4136d02b25158dc9b3d96/charts/kube-prometheus-stack/templates/grafana/dashboards-1.14/apiserver.yaml#L6

If we’re concerned about a user below k8s 1.20 wanting to use a different endpoint, we could expose this field as something that can be overridden instead, which is how Monitoring V2 does it.

But I'm not sure of a use case where a user would want to override this value.

cbron: I think the logic is fine, and it was hardcoded before, I'm more worried about the literal semver compare erroring out. But we can try this method out for now. What if we added an override first and then had this `if` as an `else if`.

aiyengar2: @cbron, instead of if / else if, I just added a default flag. 

If the user provides an override on deploying Monitoring V1 with `exporter-kubelets.kubelet.serviceMonitor.resourcePath` set to their value, it will be used instead of `/metrics/resource' or `/metrics/resource/v1alpha1`.

cbron (approved):

PR #412, opened 2021-02-12
[DNM-2.5.7] system charts for gke-operator

Klaven:
these are the system charts for the gke-operator

PR #420, opened 2021-02-26
[2.5.7] Revised addition of GKE operator chart

cmurphy:
This is a revision of #412 that incorporates the feedback from that pull request and updates the CRD to align with the current CRD in https://github.com/rancher/gke-operator/pull/3. The structure and authorship of the author's original commits are not altered, only an additional commit is added.


charts/rancher-gke-operator/1.0.000/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+name: rancher-gke-operator
+description: A Helm chart for provisioning GKE clusters
+home: https://github.com/rancher/gke-operator
+sources:
+  - "https://github.com/rancher/gke-operator"
+version: 1.0.000
+appVersion: v1.0.0
```
cbron: Lets start with app version 1.0.1 and version 1.0.100. Only because technically `1.0.000` is invalid semver, so its a weak point in our triple-zero system, even though it likely won't cause any failures.

charts/rancher-gke-operator/1.0.000/templates/clusterrole.yaml
```diff
@@ -0,0 +1,12 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: gke-operator
+  namespace: cattle-system
+rules:
```
cbron: No secrets in here yet, because no cloud creds yet ? We'll need to remember that.

charts/rancher-gke-operator/1.0.000/values.yaml
```diff
@@ -0,0 +1,11 @@
+global:
+  systemDefaultRegistry: ""
+
+gkeOperator:
+  image:
+    repository: rancher/gke-operator
+    tag: v1.0.0
```
cbron: Bump this to 1.0.1 to match chart.

charts/rancher-gke-operator/1.0.000/Chart.yaml
rmweir: @cbron app version usually just matches the operator version.

charts/rancher-gke-operator/1.0.000/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.7
```
rmweir: Min should be `2.5.7-rc1`, I believe the `rc1` is arbitrary but it needs some sort of pre-release so it works on those.

charts/rancher-gke-operator/1.0.000/templates/clusterrole.yaml
rmweir: I do think secrets are being read at least already, right?

charts/rancher-gke-operator/1.0.000/values.yaml
rmweir: @cbron we'll probably have rc images first, also why start at 1.0.1?

rmweir (changes requested):

charts/rancher-gke-operator/1.0.000/values.yaml
cbron: Because of the leading zero's being invalid semver. Not supposed to do 000. 
So this tag can match the appVersion in Chart.yaml, and we would just cut rc's on that version until we are done.

charts/rancher-gke-operator/1.0.000/Chart.yaml
cmurphy: Okay so what's the consensus? `version: 1.0.100` and `appVersion: v1.0.0` ?

charts/rancher-gke-operator/1.0.000/templates/clusterrole.yaml
cmurphy: Not currently but they will be shortly, I can add that back since we know it will be needed

charts/rancher-gke-operator/1.0.000/Chart.yaml
cbron: Ideally we keep all three together, all at 1.0.1XX

cmurphy:
Changed the version and appVersion and image tag, also added back the secret perms to the clusterrole including the create verb in light of this comment https://github.com/rancher/gke-operator/pull/3#pullrequestreview-599809746

charts/rancher-gke-operator/1.0.100/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.7
```
cbron: Hmm I thought this was right before. Should be 2.5.7-alpha1. Ex: https://github.com/rancher/system-charts/blob/dev-v2.5/charts/rancher-eks-operator/1.0.100/questions.yaml

cmurphy: Fixed

charts/rancher-gke-operator/1.0.100/values.yaml
```diff
@@ -0,0 +1,11 @@
+global:
+  systemDefaultRegistry: ""
+
+gkeOperator:
+  image:
+    repository: rancher/gke-operator
+    tag: v1.0.1
```
thedadams:
```suggestion
    tag: v1.0.1-rc1
```

charts/rancher-gke-operator/1.0.100/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+name: rancher-gke-operator
+description: A Helm chart for provisioning GKE clusters
+home: https://github.com/rancher/gke-operator
+sources:
+  - "https://github.com/rancher/gke-operator"
+version: 1.0.100
+appVersion: v1.0.1
```
thedadams:
```suggestion
appVersion: v1.0.1-rc1
```

thedadams (changes requested):

charts/rancher-gke-operator/1.0.100/values.yaml
cmurphy: done

charts/rancher-gke-operator/1.0.100/Chart.yaml
cmurphy: done

cbron (commented):
lgtm, waiting for 2.5.6 release

thedadams (approved):

rmweir (approved):
LGTM

cbron (approved):

PR #421, opened 2021-03-02
[v2.5.6] Release system-charts

ryansann:


luthermonson (dismissed):

charts/rancher-eks-operator/1.0.600/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.6-alpha1
```
deniseschannon, line 1: Do older versions need a rancher max version?

charts/rancher-monitoring/v0.1.5/questions.yml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.4.9-rc1
```
aiyengar2, line 1: Related: https://github.com/rancher/system-charts/pull/422#discussion_r585915174

deniseschannon: Same comment as within 2.4 check.

Shouldn't this be 2.4.14-rc1 and there be a max version of the monitoring version v0.1.4, so there is only 1 monitoring chart for k8s 1.16?

charts/rancher-monitoring/v0.2.1/questions.yml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.4-rc1
```
deniseschannon: Shouldn't this be 2.5.6-rc1 and the monitoring v0.2.0 should have a max Rancher version so only 1 Rancher version is available for this line of monitoring?

deniseschannon (changes requested):
Why are there changes in logging 0.2.3 when logging 0.2.3 is already released?

charts/rancher-monitoring/v0.1.5/questions.yml
aiyengar2: https://github.com/rancher/system-charts/pull/423

aiyengar2: To clarify based on offline discussion, there will still be 2 charts for Monitoring in k8s 1.16 even after the above PR is resolved: 0.1.5 and 0.2.1

However, currently 0.1.4 and 0.2.0 are also visible in 2.5.6 for k8s 1.16 users, which is what the above PR resolves by adding a rancher_max_version to those charts.

charts/rancher-eks-operator/1.0.600/questions.yaml
cbron: It picks the latest one, so not really. cc @rmweir

charts/rancher-istio/1.5.901/questions.yaml
```diff
@@ -0,0 +1,3 @@
+labels:
+  rancher.istio.v1.5.901: 1.5.9
+rancher_min_version: 2.4.5-rc1
```
cbron, line 3: We don't need to bump this and cap the previous version ? I'm not actually familiar with changes. Typically unless its security or a very bad bug we don't restrict.

cbron (commented):
1 question, needs logging fix, then looks ok

deniseschannon (dismissed):

charts/rancher-istio/1.5.901/questions.yaml
deniseschannon: I think it depends on the chart type as whether or not you can pick your own version. 

Does Istio allow you to pick what chart you want to deploy?

cbron: Yes

deniseschannon: I've approved the changes for the release, but we could discuss how much we want to limit Istio per release. 

In the past, we've limited to only 1 version based on the support matrix but not sure how much that matters/impacts for Istio.

cbron: Ok, I was unaware of the support matrix limit. I had previously thought we left multiple versions available if they were just patch or minor version changes. 
This change is not an upstream version change, just a bug on our side: https://github.com/rancher/rancher/issues/30964
Given that, I kinda think we should cap the last one.

cbron: We have decided to cap 1.5.900, putting PR for that in now.

PR #422, opened 2021-03-02
[v2.4.14] Release system-charts

ryansann:


luthermonson (dismissed):

aiyengar2 (dismissed):
LGTM

cbron (dismissed):
lgtm, 2 new versions of monitoring, and the 0.2.1 is minimum of Rancher 2.4.13 and K8s >=1.16.
Ideally Denise approves also. Otherwise sounds like we are good to cut final RC and release this.

charts/rancher-monitoring/v0.1.5/questions.yml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.4.9-rc1
```
deniseschannon: Does this mean there is more than 1 chart available for monitoring for k8s 1.16? I'd expect that this would be a min version of 2.4.14-rc1 and there would be a Rancher max version for the v0.1.4 chart.

deniseschannon (changes requested):

charts/rancher-monitoring/v0.1.5/questions.yml
aiyengar2: Yes, at the moment a k8s 1.16+ cluster could still choose to use the v0.1.x line if they wanted to use the older version of Prometheus Operator. This is true in dev-v2.5 as well as seen here: https://github.com/rancher/system-charts/pull/421/files#r585919224

aiyengar2: https://github.com/rancher/system-charts/pull/424

deniseschannon (dismissed):

aiyengar2 (dismissed):

deniseschannon (approved):

PR #433, opened 2021-03-17
Bump eks-operator to version 1.0.700

thedadams:
In order to fix an issue for users, we would like to do an out-of-band release of eks-operator. The image and chart are built, and merging this PR would result in a release to users.

Issue:
https://github.com/rancher/rancher/issues/31612

### DIFF
--- a/charts/rancher-eks-operator/1.0.700/values.yaml
+++ b/charts/rancher-eks-operator/1.0.700/values.yaml
@@ -1,0 +1,11 @@
+global:
+  systemDefaultRegistry: ""
+
+eksOperator:
+  image:
+    repository: rancher/eks-operator
+    tag: v1.0.6
+
+httpProxy: ""
+httpsProxy: ""
+noProxy: ""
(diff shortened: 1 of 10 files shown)


charts/rancher-eks-operator/1.0.700/values.yaml
```diff
@@ -0,0 +1,11 @@
+global:
+  systemDefaultRegistry: ""
+
+eksOperator:
+  image:
+    repository: rancher/eks-operator
+    tag: v1.0.6
```
cbron: Did nothing else make it into eks-operator since 2.5.6 besides this ?

thedadams: Correct, only the fix for encoded user data made it in.

rmweir (changes requested):
Can the last two commits be combined or is there a reason to keep it this way?

thedadams:
@rmweir That was the question I had during standup today. I was wondering if that was possible because `dev-v2.5` already has the `1.0.7-rc1` commit and leaving it out of `release-v2.5` might cause conflicts later.

rmweir (approved):
LGTM

cbron (approved):

PR #438, opened 2021-03-30
add additonal trusted CA option to eks-operator

bentastic27:

### DIFF
--- a/charts/rancher-eks-operator/1.0.801/Chart.yaml
+++ b/charts/rancher-eks-operator/1.0.801/Chart.yaml
@@ -4,5 +4,5 @@ description: A Helm chart for provisioning EKS clusters
 home: https://github.com/rancher/eks-operator
 sources:
   - "https://github.com/rancher/eks-operator"
-version: 1.0.800
+version: 1.0.801
 appVersion: v1.0.8
--- a/charts/rancher-eks-operator/1.0.801/Chart.yaml
+++ b/charts/rancher-eks-operator/1.0.801/Chart.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+name: rancher-eks-operator
+description: A Helm chart for provisioning EKS clusters
+home: https://github.com/rancher/eks-operator
+sources:
+  - "https://github.com/rancher/eks-operator"
+version: 1.0.800
+appVersion: v1.0.8
--- a/charts/rancher-eks-operator/1.0.801/questions.yaml
+++ b/charts/rancher-eks-operator/1.0.801/questions.yaml
@@ -1,0 +1,1 @@
+rancher_min_version: 2.5.6-alpha1
--- a/charts/rancher-eks-operator/1.0.801/questions.yaml
+++ b/charts/rancher-eks-operator/1.0.801/questions.yaml
@@ -1,1 +1,1 @@
-rancher_min_version: 2.5.6-alpha1
+rancher_min_version: 2.5.10-alpha1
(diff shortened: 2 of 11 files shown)


StrongMonkey (dismissed):
LGTM, But you will also need to make changes in rancher to pass down values.yaml option. For example, https://github.com/rancher/rancher/blob/b44c2232a71b5004f0de66833ed111540f11c6c7/pkg/controllers/management/clusteroperator/utils.go#L189

charts/rancher-eks-operator/1.0.800/templates/deployment.yaml
```diff
@@ -25,3 +25,17 @@ spec:
           value: {{ .Values.httpsProxy }}
         - name: NO_PROXY
           value: {{ .Values.noProxy }}
+{{- if .Values.additionalTrustedCAs }}
+        volumeMounts:
+        - mountPath: /etc/ssl/certs/ca-additional.pem
+          name: tls-ca-additional-volume
+          subPath: ca-additional.pem
```
thedadams: Should the `subPath` from the secret be configurable in the `values.yaml`?

charts/rancher-eks-operator/1.0.800/templates/deployment.yaml
```diff
@@ -25,3 +25,17 @@ spec:
...
+          readOnly: true
+{{- end }}
+{{- if .Values.additionalTrustedCAs }}
+      volumes:
+      - name: tls-ca-additional-volume
+        secret:
+          defaultMode: 0400
+          secretName: tls-ca-additional
```
thedadams: Should the `secretName` be configurable in the `values.yaml`?

rmweir (dismissed):
LGTM, but the changes to rancher/rancher mentioned by @StrongMonkey need to be made before this can move to test.

thedadams (changes requested):
Since version 1.0.800 has been released, I believe that the 1.0.800 folder needs to be copied to 1.0.801 and these changes made there (as well as the corresponding version bumps).

thedadams (dismissed):

rmweir (dismissed):
As long as the min version is intended, LGTM.

charts/rancher-eks-operator/1.0.801/Chart.yaml
```diff
@@ -4,5 +4,5 @@ description: A Helm chart for provisioning EKS clusters
 home: https://github.com/rancher/eks-operator
 sources:
   - "https://github.com/rancher/eks-operator"
-version: 1.0.800
+version: 1.0.801
```
cbron, line 7: Are we going to tag a 1.0.9 for 2.5.10 ?

charts/rancher-eks-operator/1.0.801/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.6-alpha1
```
cbron: This should be 2.5.10 and last chart should max at 2.5.9

charts/rancher-eks-operator/1.0.801/Chart.yaml
thedadams: No, I don't believe so.

charts/rancher-eks-operator/1.0.801/questions.yaml
thedadams: This is done.

aiyengar2 (approved):
LGTM

cbron (approved):

PR #440, opened 2021-04-01
Monitoring upgrade nginx

rmweir:


nickgerace (dismissed):

charts/rancher-monitoring/v0.2.2/questions.yml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.6-rc1
```
cbron: @aiyengar2 should we max out 0.2.1 to 2.5.6 ?

aiyengar2: We need `charts/rancher-monitoring/v0.2.1/questions.yml` to have `rancher_max_version: 2.5.7` and this questions.yml to have `rancher_min_version: 2.5.8-rc1` right?

Since we are releasing this chart in 2.5.8, but the previous chart was the latest till 2.5.7

aiyengar2 (changes requested):
Need to add questions.yml update to v0.2.1 as well, but otherwise LGTM

rmweir:
@aiyengar2 ik we've been doing this but if they already have a version of the catalog with that chart version then the max won't get updated for v0.2.1 right?

cbron:
The max should get updated if they have internet access, as it's just part of the charts that will sync. But if they have 0.2.1 installed then that will stay as-is. They just can't downgrade to it if they upgrade to 0.2.2.

charts/rancher-monitoring/v0.2.2/questions.yml
cbron: Given the change, I think we should max old one, but do we need min on this one at 2.5.8 and not 2.5.6 @aiyengar2 ?

aiyengar2:
@rmweir I think that makes sense though. We only want to prevent new users of Rancher in the new Rancher version from seeing older versions of the Monitoring chart because they should ideally be on the latest.

charts/rancher-monitoring/v0.2.2/questions.yml
```diff
@@ -1 +1 @@
-rancher_min_version: 2.5.6-rc1
+rancher_min_version: rancher_min_version: 2.5.8-rc1
```
aiyengar2: seems like a typo

aiyengar2 (changes requested):
typo, once resolved this should be good to go

charts/rancher-monitoring/v0.2.2/questions.yml
rmweir: Yep, fixed!

cbron (approved):

aiyengar2 (approved):
LGTM! Ship it

PR #442, opened 2021-04-07
Add required quotes for GKE CRD annotation

cmurphy:
See the note in the helm documentation[1]. The CRD was getting
uninstalled when the workload was deleted, causing an issue with
redeploying it. Adding the quotes to the annotation should fix it and
make it consistent with other charts in this repo.

[1] https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource


cmurphy:
Not needed

PR #444, opened 2021-04-12
Update charts missing rancher version min/max

pennyscissors:
Issue: https://github.com/rancher/rancher/issues/31568

Updated charts missing a min/max to use min `rancher_min_version: 2.3.0-rc1` and `rancher_max_version: 2.5.7` (excluding latest since we don't want to set a max in those).

### DIFF
--- a/charts/rancher-external-dns/v0.0.1/questions.yaml
+++ b/charts/rancher-external-dns/v0.0.1/questions.yaml
@@ -1,6 +1,2 @@
 rancher_min_version: 2.3.0-rc1
 rancher_max_version: 2.5.7
-categories:
-  - Logging
-
-questions:
--- a/charts/rancher-monitoring/v0.0.1/questions.yml
+++ b/charts/rancher-monitoring/v0.0.1/questions.yml
@@ -1,1 +1,2 @@
+rancher_min_version: 2.1.0-rc1
 rancher_max_version: 2.2.99
--- a/charts/rancher-eks-operator/1.0.100/questions.yaml
+++ b/charts/rancher-eks-operator/1.0.100/questions.yaml
@@ -1,1 +1,2 @@
 rancher_min_version: 2.5.0-alpha1
+rancher_max_version: 2.5.7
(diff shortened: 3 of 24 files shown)


charts/rancher-monitoring/v0.0.1/questions.yml
```diff
@@ -1 +1,2 @@
+rancher_min_version: 2.1.0-rc1
```
pennyscissors, line 1: These old monitoring charts have a max under `2.3.0-rc1` so I used min `2.1.0-rc1` because it's the oldest 2.x tag I found.

charts/rancher-external-dns/v0.0.1/questions.yaml
```diff
@@ -0,0 +1,2 @@
+rancher_min_version: 2.3.0-rc1
```
cbron, line 1: Why'd we pick 2.3.0 here, is that when it was introduced ?

charts/rancher-logging/0.0.1/questions.yml
```diff
@@ -1,4 +1,6 @@
+rancher_min_version: 2.3.0-rc1
```
cbron, line 1: Same question here, I think logging goes back further. Guess neither really matter since 2.3.0 is last supported version.

charts/rancher-monitoring/v0.1.5/questions.yml
```diff
@@ -1 +1,2 @@
 rancher_min_version: 2.5.6-rc1
+rancher_max_version: 2.5.7
```
cbron: Don't we still want this one ?

pennyscissors: Right, I'll remove this one.

charts/rancher-logging/0.0.1/questions.yml
pennyscissors: I chose 2.3.0-rc1 for these because 2.3 is the oldest rancher version we still support in some way (security stuff only afaik), but I can change them to something even older if you prefer.

cbron: Looks like its from 2.2, but we'll keep 2.3 since they should have upgraded by now. Same for DNS.

cbron (commented):
lgtm. Can you run a test with and without this and diff that.

paynejacob (approved):

pennyscissors:
Closing in favor of this [issue](https://github.com/rancher/rancher/issues/33037) as these changes may no longer be relevant.

PR #450, opened 2021-04-29
[v2.5.8] Release system-charts

pennyscissors:


charts/rancher-eks-operator/1.0.800/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.6-alpha1
```
cbron, line 1: @PennyScissors should the former chart be maxing version to be compatible with your new images.txt system ?

charts/rancher-gke-operator/1.0.100/values.yaml
```diff
@@ -0,0 +1,11 @@
+global:
+  systemDefaultRegistry: ""
+
+gkeOperator:
+  image:
+    repository: rancher/gke-operator
+    tag: v1.0.1-rc12
```
cbron: This will need to be `v1.0.1` before we merge

charts/rancher-eks-operator/1.0.800/questions.yaml
pennyscissors: Yes, if we don't add a max the old version, all the images in that chart will be added to the image list in future releases once my changes are in.

charts/rancher-gke-operator/1.0.100/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.7-alpha1
```
PennyScissors, line 1: Same for this chart, the previous version needs a max.

charts/rancher-gke-operator/1.0.100/values.yaml
pennyscissors: cc @cmurphy

cbron: To be clear its correct for now. GKE still has open issues. We just need to remind ourselves to cut one later.

superseb: Is this good with `v1.0.1-rc13` or are changing it still?

charts/rancher-gke-operator/1.0.100/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+name: rancher-gke-operator
+description: A Helm chart for provisioning GKE clusters
+home: https://github.com/rancher/gke-operator
+sources:
+  - "https://github.com/rancher/gke-operator"
+version: 1.0.100
+appVersion: v1.0.1-rc13
```
cbron: I can't comment on the old thread, but we can't merge this until we release GKE image. GKE is not finished yet.

cbron (changes requested):

charts/rancher-gke-operator/1.0.100/questions.yaml
cbron: @cmurphy Shouldn't this be: 
```suggestion
rancher_min_version: 2.5.8-alpha1
```

cmurphy: @cbron yes it should

cmurphy: https://github.com/rancher/system-charts/pull/455

SheilaghM (approved):

paynejacob (approved):

cbron (approved):

PR #451, opened 2021-04-30
Bump gke-operator to RC12

cmurphy:


thedadams (dismissed):

rmweir (commented):
Are the CRDs any different after the updates to norman tags?

cmurphy:
@rmweir not for the norman tags but it is for the new labels field, thanks for catching

rmweir (dismissed):
LGTM

cbron (approved):

PR #452, opened 2021-04-30
[dev-v2.5] Set max for rancher-eks-operator 1.0.700

pennyscissors:


cbron (approved):
lgtm, cc @rmweir and @thedadams we have to do this going forward so old ones dont get airgapped

PR #454, opened 2021-05-04
Istio mirror images

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/31598

**Dependent PR**
https://github.com/rancher/image-mirror/pull/91

**Problem**
Images that are mirror from an upstream repository into a rancher dockerhub repository need to be mirrored into a rancher dockerhub respository that has "mirrored" in the title

**Solution**
Mirror the istio chart images to a dockerhub repo that has mirrored and update chart to use those images


paynejacob (dismissed):

rmweir (commented):
Should the `rancher_min_version` be updated to 2.5.8-rc1 and the `rancher_max_version` of the previous version of the istio chart be set to 2.5.7? @cbron thoughts? Export script will be changed in 2.6 to extract images from all compatible versions so something like this would cause non mirrored images to remain in rancher-images and rancher-images-sources text files.

brendarearden:
@rmweir you are correct. I have updated the min/max version to reflect 2.5.7 for the max and 2.5.8-rc1 as the newest version's min.

cbron:
This may actually not be needed, holding off on review for now.

paynejacob (dismissed):

rmweir (dismissed):
LGTM

charts/rancher-istio/1.5.902/questions.yaml
```diff
@@ -1,3 +1,3 @@
 labels:
   rancher.istio.v1.5.902: 1.5.9
-rancher_min_version: 2.4.5-rc1
+rancher_min_version: 2.5.8-rc1
```
cbron: This could be 2.6 ~but is actually fine for now since we may do 2.5.x anyway~ need to handle anyway

cbron (dismissed):
I didn't check that these all exist. You should pull them to double check.

charts/rancher-istio/1.5.902/questions.yaml
brendarearden: @cbron does that mean I Should set the max version on the 1.5.901chart to 1.5.10?

charts/rancher-istio/1.5.902/Chart.yaml
```diff
@@ -14,4 +14,4 @@ name: rancher-istio
 sources:
 - http://github.com/istio/istio
 tillerVersion: '>=2.7.2-0'
-version: 1.5.901
+version: 1.5.902
```
cbron: Because this is 2.6, I think you have to make this 1.5.920. That way several versions in 2.5 can still use 1.5.902-1.5.919

cbron (commented):
In 2.6: we will make this 1.5.920 with min of 2.6.0. It will effectively be a new change.
In 2.5: This will effectively be a backport, we can make a "new" version down there of 1.5.902. That one will have to be forwardported to 2.6 also so the min will just be 2.5.15 (or whatever version its introduce in) with no max.

charts/rancher-istio/1.5.902/questions.yaml
cbron: For now just leave it as-is, because it needs to be available in 2.6. Until we make 1.5.902 in 2.5.

charts/rancher-istio/1.5.902/Chart.yaml
brendarearden: Updated to be 1.5.920 so that our 2.5 branch can continue to have additional fixes - I went ahead and left it as a separate commit, but will merge the commits once review is completed and PR is approved

charts/rancher-istio/1.5.901/questions.yaml
```diff
@@ -1,3 +1,5 @@
 labels:
   rancher.istio.v1.5.901: 1.5.9
 rancher_min_version: 2.4.5-rc1
+rancher_max_version: 2.5.10
```
cbron: nit: we've been putting this to 2.5.99, but effectively it doesn't matter

charts/rancher-istio/1.5.920/values.yaml
```diff
@@ -0,0 +1,698 @@
...
+    # flags to true.
+    gatewayName: ingressgateway
+    # enableHttps will add port 443 on the ingress.
+    # It REQUIRES that the certificates are installed  in the
+    # expected secrets - enabling this option without certificates
+    # will result in LDS rejection and the ingress will not work.
+    enableHttps: false
+
+  curl:
+    repository: rancher/mirrored-pstauffer-curl
```
cbron, line 200: tag got deleted here

cbron (changes requested):

charts/rancher-istio/1.5.920/Chart.yaml
```diff
@@ -14,4 +14,4 @@ name: rancher-istio
 sources:
```
cbron, line 14: Odd that the original folder here was 1.5.902 instead of 1.5.901, guessing that was folder name of copy, and that would never go public.

cbron (approved):
lgtm

charts/rancher-istio/1.5.920/Chart.yaml
brendarearden: Correct - in the first commit I did was a copy of 1.5.901 and then rename it to 1.5.902. In the second commit I made the changes I needed to and that included the rename to 1.5.920 so that is why its showing up that way.

PR #458, opened 2021-05-05
Update gke-operator to v1.0.1

cmurphy:


cbron (approved):
Technically will be removed, but fine for now for scanning purposes.

PR #459, opened 2021-05-07
Cut monitoring v0.2.3 and use mirrored image

jiaqiluo:
**This PR is closed, here is why: https://github.com/rancher/system-charts/pull/459#issuecomment-841474381**

Issue: https://github.com/rancher/rancher/issues/31871 

also address this issue: https://github.com/rancher/rancher/issues/32693

Mirroring images is already handled by the PR: https://github.com/rancher/image-mirror/pull/96


charts/rancher-monitoring/v0.2.3/values.yaml
```diff
@@ -251,7 +251,7 @@ grafana:
   level: cluster
   apiGroup: "monitoring.coreos.com"
   image:
-    repository: rancher/grafana-grafana
+    repository: rancher/mirrored-grafana-grafana
     tag: 7.1.5
     tool:
       repository: rancher/prometheus-auth
```
cbron, line 257: Is this one rancher generated ?

charts/rancher-monitoring/v0.2.3/values.yaml
```diff
@@ -311,7 +311,7 @@ prometheus:
           fieldPath: status.podIP
   apiGroup: "monitoring.coreos.com"
   image:
-    repository: rancher/prom-prometheus
+    repository: rancher/mirrored-prom-prometheus
     tag: v2.18.2
     auth:
       repository: rancher/prometheus-auth
```
cbron, line 317: same

charts/rancher-monitoring/v0.2.3/values.yaml
```diff
@@ -142,7 +142,7 @@ exporter-node:
   enabled: false
   apiGroup: "monitoring.coreos.com"
   image:
-    repository: rancher/prom-node-exporter
+    repository: rancher/mirrored-prom-node-exporter
     tag: v1.0.1
   nodeSelectors: []
   resources:
```
cbron, line 148: is wmi_exporter below rancher generated ?

charts/rancher-monitoring/v0.2.3/values.yaml
```diff
@@ -391,8 +391,8 @@ operator-init:
   enabled: false
   apiGroup: "monitoring.coreos.com"
   image:
-    repository: rancher/istio-kubectl
-    tag: 1.5.10
+    repository: rancher/kubectl
+    tag: v1.20.2
   resources:
     limits:
       cpu: 20m
```
cbron, line 398: same for webhook receiver below

charts/rancher-monitoring/v0.2.3/values.yaml
jiaqiluo: Yes, Rancher generates the image https://github.com/rancher/wmi_exporter-package

charts/rancher-monitoring/v0.2.3/values.yaml
jiaqiluo: yes, https://github.com/rancher/prometheus-auth

charts/rancher-monitoring/v0.2.3/values.yaml
jiaqiluo: yes https://github.com/rancher/prometheus-auth

charts/rancher-monitoring/v0.2.3/values.yaml
jiaqiluo: yes https://github.com/rancher/webhook-receiver

charts/rancher-monitoring/v0.2.3/questions.yml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.8-rc1
```
cbron: This should be 2.5.9-rc1, and last chart version should have a max of 2.5.8. We should also make this change in 2.6 branch with equivalent min/max versions.

cbron (commented):
lgtm besides one change, needs arvind approval also

charts/rancher-monitoring/v0.2.3/Chart.yaml
```diff
@@ -0,0 +1,17 @@
...
+description: Provides monitoring for Kubernetes which is maintained by Rancher 2.
+engine: gotpl
+maintainers:
+- name: thxCode
+  email: frank@rancher.com
+name: rancher-monitoring
+sources:
+- https://github.com/coreos/prometheus-operator
+version: 0.2.3
+appVersion: 0.2.3
```
aiyengar2, line 11: Can we split the base copy and modification changes into separate commits? I'm assuming all you changed was the version / appVersion and will be modifying the questions.yml based on Caleb's comment in https://github.com/rancher/system-charts/pull/459#discussion_r630383265, but this should be in a separate commit.

The base copy commit should generally be the equivalent of doing `cp -R charts/rancher-monitoring/v0.2.2 charts/rancher-monitoring/v0.2.3`.

aiyengar2 (commented):
Need to address Caleb's comment on the question.yml and split base copy changes into 2 commits, but otherwise LGTM.

Can approve / merge once those changes are addressed.

aiyengar2 (changes requested):
Forgot one more thing; we need to replicate these changes to release a new v0.1.x chart as well since we still support k8s 1.15 clusters in Rancher 2.4, which are not eligible to use this chart.

@jiaqiluo can you add that to this PR as well as a separate set of commits? ~No need for another PR.~ You need another PR if and only if this needs to be backported to Rancher 2.4 since v0.1.x doesn't need to be added to 2.5 as 2.5 does not need to support k8s 1.15

charts/rancher-monitoring/v0.2.3/questions.yml
jiaqiluo: @cbron @aiyengar2  But the v0.2.2 chart does not set the rancher_max_version, and we can't change the released chart, right?

aiyengar2: Generally we do e.g. https://github.com/rancher/system-charts/commit/23569d0ee2a71a37523a9deb50f53de342a51f34.

I think the assumption is that we can set the rancher_max_version since, during an Rancher upgrade, the Helm charts are all re-downloaded. So the chart annotation gets picked up.

jiaqiluo: ok. If so I will handle that for the v0.2.2 chart. 
Update: it is tracked now at https://github.com/rancher/rancher/issues/32693

charts/rancher-monitoring/v0.2.3/Chart.yaml
jiaqiluo: done.

jiaqiluo:
> Forgot one more thing; we need to replicate these changes to release a new v0.1.x chart as well since we still support k8s 1.15 clusters in Rancher 2.4, which are not eligible to use this chart.
> 
> @jiaqiluo can you add that to this PR as well as a separate set of commits? No need for another PR.

@cbron @aiyengar2 
~Do we really need to backport this change to the v0.1.x line?~ 

~Here are the version requirements for v0.1.4 and v0.1.5~
```
v0.1.4
rancher_min_version: 2.4.9-rc1
rancher_max_version: 2.5.5

v0.1.5
rancher_min_version: 2.5.6-rc1
```

~Will it be confusing if we cut a v0.1.6 which is available on rancher v2.4.x?~ 
~Or the v0.1.6 should has rancher_min_version = 2.5.9-rc1 ?~

**Update:
Oh, I think I should make the new version on the dev-2.4 branch if needed. I was looking at the dev-v2.5 branch wrongly.**

aiyengar2 (approved):
LGTM

charts/rancher-monitoring/v0.2.2/questions.yml
```diff
@@ -1 +1,2 @@
 rancher_min_version: 2.5.8-rc1
+rancher_max_version: 2.5.9
```
aiyengar2:
```suggestion
rancher_max_version: 2.5.8
```

aiyengar2 (changes requested):
I approved too soon! The rancher_max_version needs to be changed.

charts/rancher-monitoring/v0.2.2/questions.yml
jiaqiluo: fixed.

charts/rancher-monitoring/v0.2.2/questions.yml
```diff
@@ -1 +1,2 @@
 rancher_min_version: 2.5.8-rc1
+rancher_max_version: 2.5.8
```
jiaqiluo, line 2: We limit chart v0.2.2 to be available in Rancher v2.5.8 only

Rancher reads the above values as 
```
2.5.8-rc1 <= target-version <= 2.5.8
```

aiyengar2 (dismissed):
LGTM

charts/rancher-monitoring/v0.2.3/questions.yml
cbron: You are right that its a little weird, but its common practice for us. I don't think you really need an issue to track it, but we can leave that one.

charts/rancher-monitoring/v0.2.2/questions.yml
cbron, line 2: Code looks good, but few more things: 
* based on convo earlier we only need to do mirroring in 2.6, so I can move this to point at dev-v2.6 branch instead
* Do we need to adjust the other chart like arvind said ? 
* Lets squash to only 2 commits, 1 copy and 1 for changes

jiaqiluo: > Do we need to adjust the other chart like arvind said ?

Do you mean the v0.1.x chart that Arvind mentioned [here](https://github.com/rancher/system-charts/pull/459#pullrequestreview-657235561)?  I think we do not need to do it since we will only release the new monitoring chart in the monitoring 0.2.x line in Rancher v2.6

(NOTE: I will move this PR to the dev-v2.6 branch and squash commits after we make a decision on the above question.)

cbron: Ok so 2.6 only, no old chart. I will change base branch, just need to squash commits.

jiaqiluo:
This PR is closed because:
1. the new chart will go to the dev-v2.6 branch only
2. it is too complex to fix the conflicts because of the change of the base image

https://github.com/rancher/system-charts/pull/460 is the final PR going to the dev-v2.6

PR #463, opened 2021-05-20
Update gke-operator to v1.0.2-rc1

cmurphy:
https://github.com/rancher/rancher/issues/32585

### DIFF
--- a/charts/rancher-gke-operator/1.0.200/Chart.yaml
+++ b/charts/rancher-gke-operator/1.0.200/Chart.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+name: rancher-gke-operator
+description: A Helm chart for provisioning GKE clusters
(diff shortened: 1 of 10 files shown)


charts/rancher-gke-operator/1.0.200/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
cbron, line 1: For new charts in system-charts we do a first commit that is just a copy of old chart, and then 2nd commit that has any changes, that way you can view the real diff. Changes here would just be around version and such.

cmurphy: done

cbron (approved):

thedadams (approved):

PR #464, opened 2021-05-20
Update gke-operator to to 2.0.1-rc2

cmurphy:
https://github.com/rancher/rancher/issues/32597


cmurphy:
Closing to update this in the feature chart instead. This forwardport bug won't be able to be tested until we're on the feature chart.

cbron:
If we do end up using system-charts we'll need to ensure any updates get to here. We can close this one, but any more changes in operator should get updated in system-charts.

cmurphy:
Reopening using 1.1.1-rc1, since we're going to continue using system-charts until the feature chart conversion is ready.

cmurphy:
Reopened here https://github.com/rancher/system-charts/pull/468

PR #468, opened 2021-05-24
Update gke-operator to 1.1.1-rc1

cmurphy:
https://github.com/rancher/rancher/issues/32597

### DIFF
--- a/charts/rancher-gke-operator/1.1.100/questions.yaml
+++ b/charts/rancher-gke-operator/1.1.100/questions.yaml
@@ -1,0 +1,1 @@
+rancher_min_version: 2.5.7-alpha1
--- a/charts/rancher-gke-operator/1.1.100/questions.yaml
+++ b/charts/rancher-gke-operator/1.1.100/questions.yaml
@@ -1,1 +1,1 @@
-rancher_min_version: 2.5.7-alpha1
+rancher_min_version: 2.6.0-alpha1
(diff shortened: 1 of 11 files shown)


charts/rancher-gke-operator/1.1.100/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.7-alpha1
```
cbron: We should make this 2.6.0-alpha1

cbron (changes requested):
This chart will likely be deleted before 2.6 ships, but thats ok, its a holdover.

charts/rancher-gke-operator/1.1.100/questions.yaml
cmurphy: done

cbron (dismissed):

charts/rancher-gke-operator/1.1.100/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.6.0-alpha1
```
cbron, line 1: Oh wait, actually can you set max for last chart in questions.yaml to 2.5.99. Without that it will get included in the images.txt list.

cmurphy: done

cbron (approved):

PR #470, opened 2021-05-25
Update mirrored sonobuoy version for CIS v1

prachidamle:
https://github.com/rancher/rancher/issues/31872
https://github.com/rancher/rancher/issues/31598


charts/rancher-cis-benchmark/0.1.3/values.yaml
```diff
@@ -30,7 +30,7 @@ image:
     tag: v0.1.14
```
cbron: The chart.yaml version here is still 0.1.2. Since this is going into 2.6 you should bump the version to 0.2.0.

prachidamle: Ok, changed to 0.2.0

charts/rancher-cis-benchmark/0.1.2/questions.yaml
```diff
@@ -1 +1,2 @@
 rancher_min_version: 2.4.6-rc1
+rancher_max_version: 2.5.8
```
cbron, line 2: Hmm. We could make this 2.5.99 for when a new version gets introduced in 2.5.x, but I suppose this works for now since its in 2.6 branch and can't be affected. We need some max so that it doesn't get included in images.txt.

charts/rancher-cis-benchmark/0.2.0/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.9-alpha1
```
cbron: Should be 2.6.0-alpha1, this is going into 2.6. Anything in 2.5.x line would be 0.1.3 in the dev-v2.5. branch.

charts/rancher-cis-benchmark/0.2.0/values.yaml
```diff
@@ -30,7 +30,7 @@ image:
     tag: v0.1.14
     pullPolicy: Always
   sonobuoy:
-    repository: rancher/sonobuoy-sonobuoy
+    repository: rancher/mirrored-sonobuoy-sonobuoy
```
cbron, line 33: Is there a newer version of this ? See scanning issue.

prachidamle: I see this mirrored already - https://hub.docker.com/r/rancher/mirrored-sonobuoy-sonobuoy 
Do you mean pulling in updated upstream sonobuoy?

cbron: Talked offline, we are going to check the upstream docs to see if we can upgrade this to newer sonobuoy.

prachidamle: I created this ticket  https://github.com/rancher/rancher/issues/33335 to test CIS with upstream newer sonobuoy to make sure there is no breaking change, can we push this fix to introduce the mirroring fixes with the existing version @cbron cc @deniseschannon

cbron (approved):
Lgtm, will address sonobuoy later.

PR #471, opened 2021-05-26
change monitoring v1 chart version from v0.2.3 to v0.3.0

jiaqiluo:
As @cbron  suggests, the monitoring `v0.2.x` line should be reserved for the Rancher `v2.5.x` line. 
So we need to change v0.2.3 to v0.3.0 in the dev-v2.6 branch 

This PR does the following:
 - change the directory name from v0.2.3 to v0.3.0
 - change the `version` and `appVerson` in `Chart.yaml`
 - change the  `rancher_min_version` in `questions.yaml`

Todo after the PR is merged:

- [ ] update the monitoring chart version in existing related issues from v0.2.3 to v0.3.0


charts/rancher-monitoring/v0.2.3/questions.yml
```diff
@@ -1 +0,0 @@
-rancher_min_version: 2.5.9-rc1
```
jiaqiluo, line 1: The value of `rancher_min_version` is changed to `2.6.0-rc1`, but Git treats it as a removal because the directory name is also changed. 

There is a file creation if you scroll down on this page.

charts/rancher-monitoring/v0.3.0/questions.yml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.6.0-rc1
```
jiaqiluo, line 1: **Question:** Should the value be `2.6.0-0` in case we want this chart showing in Rancher versions like 2.6.0-alpha or so?

aiyengar2: Yes I think that makes sense.

aiyengar2 (changes requested):
@jiaqiluo hold off on this change since I think the consensus Caleb, Denise, and I reached today involved bumping the **major** version during a Rancher minor release. 

Therefore, if you want to modify this PR in advanced, the version should be 1.0.0.

jiaqiluo:
Close this PR as it is has been done by another PR.

PR #472, opened 2021-05-26
Add additional CAs to GKE chart

thedadams:
When Rancher is installed on a cluster behind a proxy, gke-operator
needs to trust the certificate from the proxy. Therefore, it is passed
to the deployment as a secret that is setup by the user as per the
documentation.

Issue:

### DIFF
--- a/charts/rancher-gke-operator/1.0.200/templates/deployment.yaml
+++ b/charts/rancher-gke-operator/1.0.200/templates/deployment.yaml
@@ -25,3 +25,15 @@ spec:
           value: {{ .Values.httpsProxy }}
         - name: NO_PROXY
           value: {{ .Values.noProxy }}
+{{- if .Values.additionalTrustedCAs }}
+        volumeMounts:
+        - mountPath: /etc/ssl/certs/ca-additional.pem
+          name: tls-ca-additional-volume
+          subPath: ca-additional.pem
+          readOnly: true
+      volumes:
+      - name: tls-ca-additional-volume
+        secret:
+          defaultMode: 0400
+          secretName: tls-ca-additional
+{{- end }}
--- a/charts/rancher-gke-operator/1.0.200/values.yaml
+++ b/charts/rancher-gke-operator/1.0.200/values.yaml
@@ -9,3 +9,4 @@ gkeOperator:
 httpProxy: ""
 httpsProxy: ""
 noProxy: ""
+additionalTrustedCAs: false


charts/rancher-gke-operator/1.0.200/templates/deployment.yaml
```diff
@@ -25,3 +25,17 @@ spec:
           value: {{ .Values.httpsProxy }}
         - name: NO_PROXY
           value: {{ .Values.noProxy }}
+{{- if .Values.additionalTrustedCAs }}
+        volumeMounts:
+        - mountPath: /etc/ssl/certs/ca-additional.pem
```
nickgerace, line 30: Is this always going to be the same path? I cannot recall if you can change the underlying distro in GKE, and if you can, if GKE overrides default paths anyway.

nickgerace (commented):
Does `packageVersion` need to be bumped?

charts/rancher-gke-operator/1.0.200/templates/deployment.yaml
thedadams: This chart is installed on the local cluster. This is the same path that Rancher uses: https://github.com/rancher/rancher/blob/master/chart/templates/deployment.yaml#L136-L145

nickgerace (approved):

cmurphy (approved):

charts/rancher-gke-operator/1.0.200/templates/deployment.yaml
cbron: This mountPath is inside the gke-operator pod, so it could be any path as long as the operator looks for it. We can keep using same one as rancher for uniformity sake. 
How does the operator pod use it from there ?

cbron (commented):
I updated PR to point to backport issue

charts/rancher-gke-operator/1.0.200/templates/deployment.yaml
thedadams: It's put in that directory because the libraries automatically look in that place for CAs. It is automatically picked up and the pod/gke-operator doesn't need to do anything else.

cbron (approved):

cbron:
This will need to be added to the 2.6 branch as well as-is, for those that upgrade from 2.5 to 2.6

thedadams:
@cbron It is added to 2.6 in v1.1.100 for those that upgrade to 2.6. Since the chart gets upgraded automatically in Rancher, I thought this would be enough.

cbron:
I thought so too, but according to our new versioning system any version that ever gets merged in old releases needs to get forwardported to all newer versions. So 2.6 should have 1.0.100, 1.0.200, and 1.1.100.

PR #474, opened 2021-05-26
Add additional trusted CAs if needed

thedadams:
When Rancher is installed on a cluster behind a proxy, the operator
needs to trust the certificate from the proxy. This is done via the
deployment and a secret the user sets up as per the documentation.

Issue:

### DIFF
--- a/charts/rancher-aks-operator/1.0.100/templates/deployment.yaml
+++ b/charts/rancher-aks-operator/1.0.100/templates/deployment.yaml
@@ -25,3 +25,15 @@ spec:
           value: {{ .Values.httpsProxy }}
         - name: NO_PROXY
           value: {{ .Values.noProxy }}
+{{- if .Values.additionalTrustedCAs }}
+        volumeMounts:
+        - mountPath: /etc/ssl/certs/ca-additional.pem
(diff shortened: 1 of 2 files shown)


charts/rancher-aks-operator/1.0.100/templates/deployment.yaml
```diff
@@ -25,3 +25,17 @@ spec:
           value: {{ .Values.httpsProxy }}
         - name: NO_PROXY
           value: {{ .Values.noProxy }}
+{{- if .Values.additionalTrustedCAs }}
+        volumeMounts:
+        - mountPath: /etc/ssl/certs/ca-additional.pem
+          name: tls-ca-additional-volume
+          subPath: ca-additional.pem
+          readOnly: true
+{{- end }}
```
brendarearden: Is this being split out into separate if statements incase we add more volume mounts or could they be in the same if  check?

thedadams: Good point. For now they could be in the same check.

mjura (approved):
LGTM

cbron (approved):

PR #475, opened 2021-05-26
Add additional CA to eks-operator

thedadams:
Issue:

### DIFF
--- a/charts/rancher-eks-operator/1.0.800/Chart.yaml
+++ b/charts/rancher-eks-operator/1.0.800/Chart.yaml
@@ -5,4 +5,4 @@ home: https://github.com/rancher/eks-operator
 sources:
   - "https://github.com/rancher/eks-operator"
 version: 1.0.800
-appVersion: v1.0.8-rc2
+appVersion: v1.0.8
--- a/charts/rancher-eks-operator/1.1.100/values.yaml
+++ b/charts/rancher-eks-operator/1.1.100/values.yaml
@@ -4,8 +4,9 @@ global:
 eksOperator:
   image:
     repository: rancher/eks-operator
-    tag: v1.0.8-rc2
+    tag: v1.1.1-rc2
 
 httpProxy: ""
 httpsProxy: ""
 noProxy: ""
+additionalTrustedCAs: false
--- a/charts/rancher-eks-operator/1.1.100/values.yaml
+++ b/charts/rancher-eks-operator/1.1.100/values.yaml
@@ -1,0 +1,11 @@
+global:
+  systemDefaultRegistry: ""
+
+eksOperator:
+  image:
+    repository: rancher/eks-operator
+    tag: v1.0.8-rc2
+
+httpProxy: ""
+httpsProxy: ""
+noProxy: ""
--- a/charts/rancher-eks-operator/1.0.100/questions.yaml
+++ b/charts/rancher-eks-operator/1.0.100/questions.yaml
@@ -1,1 +1,2 @@
 rancher_min_version: 2.5.0-alpha1
+rancher_max_version: 2.5.5
--- a/charts/rancher-eks-operator/1.0.101/questions.yaml
+++ b/charts/rancher-eks-operator/1.0.101/questions.yaml
@@ -1,1 +1,2 @@
 rancher_min_version: 2.5.0-alpha1
+rancher_max_version: 2.5.5
--- a/charts/rancher-eks-operator/1.0.400/questions.yaml
+++ b/charts/rancher-eks-operator/1.0.400/questions.yaml
@@ -1,1 +1,2 @@
 rancher_min_version: 2.5.0-alpha1
+rancher_max_version: 2.5.5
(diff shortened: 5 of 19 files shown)


charts/rancher-eks-operator/1.0.801/Chart.yaml
```diff
@@ -4,5 +4,5 @@ description: A Helm chart for provisioning EKS clusters
 home: https://github.com/rancher/eks-operator
 sources:
   - "https://github.com/rancher/eks-operator"
-version: 1.0.800
+version: 1.0.801
```
aiyengar2: Do we need a questions.yml change in both this chart and the previous chart for rancher min/max version?

charts/rancher-eks-operator/1.0.801/templates/deployment.yaml
```diff
@@ -25,3 +25,17 @@ spec:
...
+            name: tls-ca-additional-volume
+            subPath: ca-additional.pem
+            readOnly: true
+        {{- end }}
+      {{- if .Values.additionalTrustedCAs }}
+      volumes:
+      - name: tls-ca-additional-volume
+        secret:
+          defaultMode: 0400
+          secretName: tls-ca-additional
```
aiyengar2: Should this value be providable by the user? Also, should we add this to the documentation (README.md?)

aiyengar2 (changes requested):

charts/rancher-eks-operator/1.0.801/Chart.yaml
thedadams: I don't think so. Rancher auto-upgrades the chart as needed and both chart versions are compatible.

charts/rancher-eks-operator/1.0.801/templates/deployment.yaml
thedadams: That was my original thought, but this follows what is done in Rancher: https://github.com/rancher/rancher/blob/master/chart/templates/deployment.yaml#L136-L145

aiyengar2 (dismissed):
LGTM

charts/rancher-eks-operator/1.0.801/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.6-alpha1
```
rmweir: Should this be changed to the 2.5.9-alpha1?

rmweir (changes requested):

charts/rancher-eks-operator/1.0.801/questions.yaml
thedadams: The chart will work on 2.5.6+ because `additionalTrustedCAs` is false by default. Even if a user sets it to `true` when deploying the helm chart, the chart will deploy successfully as long as the secret exists as in the Rancher docs.

rmweir: Okay, should be fine. I'm not sure if we have stances on when we should let older versions of rancher use newer charts. @cbron do you have any insights on that?

rmweir (dismissed):
LGTM

charts/rancher-eks-operator/1.0.801/Chart.yaml
cbron: The problem is that now the image export script takes any chart that matches the current rancher version, so we do need to cap the old one.

charts/rancher-eks-operator/1.0.801/questions.yaml
cbron: It depends on the chart. If there are no UI/API changes then it can go into previous rancher versions, but to be safe we can always min the new one.

cbron: I just realized this is 2.6. This whole chart needs to get bumped to 1.1.100. With min of 2.6-alpha1, and the old chart should get a max of 2.5.99 (only in dev-v2.6 branch).

thedadams: Done.

charts/rancher-eks-operator/1.0.800/Chart.yaml
```diff
@@ -5,4 +5,4 @@ home: https://github.com/rancher/eks-operator
 sources:
   - "https://github.com/rancher/eks-operator"
 version: 1.0.800
-appVersion: v1.0.8-rc2
+appVersion: v1.0.8
```
cbron, line 8: This is ok, since this is what was already in 2.5 release branch. We should ensure the changes here are only differences between those two.

charts/rancher-eks-operator/1.1.100/values.yaml
```diff
@@ -4,8 +4,9 @@ global:
 eksOperator:
   image:
     repository: rancher/eks-operator
-    tag: v1.0.8-rc2
+    tag: v1.0.8
```
cbron: We have rc for 1.1.1 already, should be updated in app-version too. We should really tag a new RC, there are newer commits now.

cbron (commented):
small change then good

charts/rancher-eks-operator/1.1.100/values.yaml
thedadams: Done

cbron (dismissed):

cbron (approved):
lgtm, we added the max to old charts so they don't get picked up in images.txt

PR #478, opened 2021-05-28
Forwardports gke-operator v1.0.200

thedadams:
The new versioning changes require that versions in dev-v2.5 also exist
in dev-v2.6. This change copies the files from dev-v2.5.

2.5 PR:
https://github.com/rancher/system-charts/pull/472

Issue:

### DIFF
--- a/charts/rancher-gke-operator/1.0.200/questions.yaml
+++ b/charts/rancher-gke-operator/1.0.200/questions.yaml
@@ -1,0 +1,1 @@
+rancher_min_version: 2.5.8-patch1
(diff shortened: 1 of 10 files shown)


cbron:
Adding arvind as reviewer just to confirm this follows the expectations he wrote up

charts/rancher-gke-operator/1.0.200/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.8-patch1
```
cbron, line 1: Oh wait, I forgot GKE is getting remove entirely from 2.6. This PR likely isn't needed. It would only be needed if we keep it in system-charts. Sorry for double work.

cbron:
Closing for now, we can re-open if we stick to system-charts

PR #481, opened 2021-06-15
Bump gke-operator image to 1.1.1-rc3

cmurphy:


cmurphy:
@cbron is this what we want for dev-v2.5? Do we need the same for dev-v2.6?

cbron (approved):
Yes this is correct, although I'm somewhat surprised there aren't any additional chart changes.

We need to make sure a copy of this 1.0.200 chart ends up in dev-v2.6 branch also, but we don't need to introduce anything else in 2.6 system-charts. What we are doing is allowing them to use this 2.5 version in 2.5 and then upgrade to 2.6 and stay on the version, but once there they should be able to upgrade to the 2.6 version from feature-charts.
I suppose the auto-deploy functionality may negate some of that process, but still good to follow.

cmurphy:
> although I'm somewhat surprised there aren't any additional chart changes.

There haven't been any CRD changes between 1.0 and 1.1 so this is correct.

cmurphy:
v2.6 https://github.com/rancher/system-charts/pull/482

PR #483, opened 2021-06-15
Forward-port gke-operator 1.0.200

cmurphy:
Adds 1.0.200 package for gke-operator from dev-v2.5.


cbron (approved):
lgtm, just don't forget to update over time. I don't think the questions.yaml should matter, especially since the operator image is the same as the newer feature chart version, so it will be airgapped.

PR #493, opened 2021-07-06
Add eks-operator 1.0.1000

thedadams:
Related issue:
https://github.com/rancher/rancher/issues/33172


cmurphy (dismissed):

rmweir (dismissed):
LGTM

charts/rancher-eks-operator/1.0.1000/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+name: rancher-eks-operator
+description: A Helm chart for provisioning EKS clusters
+home: https://github.com/rancher/eks-operator
+sources:
+  - "https://github.com/rancher/eks-operator"
+version: 1.0.1000
+appVersion: v1.0.10-rc1
```
cbron, line 8: confused about this, we have: 
1.0.800 in release: https://github.com/rancher/system-charts/tree/release-v2.5/charts/rancher-eks-operator
1.0.801 in dev already: https://github.com/rancher/system-charts/tree/dev-v2.5/charts/rancher-eks-operator
and now we are adding 1.0.1000 ? Why aren't we using 1.0.801.

cbron (changes requested):

charts/rancher-eks-operator/1.0.1000/Chart.yaml
thedadams: I was trying to keep the chart and image versions in line. There was the accidental inclusion of v1.0.9.

We can delete the v1.0.9 tag and re-use it. Or stay with v1.0.10.

cbron (approved):
lgtm, already has min of 2.5.10.

PR #497, opened 2021-07-07
Update gke-operator image to v1.0.2-rc2

cmurphy:


thedadams (dismissed):

rmweir (dismissed):
LGTM

charts/rancher-gke-operator/1.0.200/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.8-patch1
```
cbron: Does this need to be 2.5.9 for rancher changes ? Will Rancher 2.5.8 handle this version.

cmurphy: It might have to be 2.5.9 because of https://github.com/rancher/rancher/pull/33473

thedadams (dismissed):

thedadams (approved):

charts/rancher-gke-operator/1.0.200/questions.yaml
```diff
@@ -1 +1 @@
-rancher_min_version: 2.5.8-patch1
+rancher_min_version: 2.5.9-rc5
```
cbron, line 1: Going with rc5 here since that is version of rancher its compatible with.

cbron (approved):
lgtm. This will bump anything else in dev-v2.5 up to 1.0.300+

PR #496, opened 2021-07-07
[v2.5] Bump gke-operator to 1.1.1-rc7

cmurphy:
https://github.com/rancher/rancher/issues/33458

PR #498, opened 2021-07-07
Bump gke-operator to 1.1.1-rc7

cmurphy:
https://github.com/rancher/rancher/issues/33406

### DIFF
--- a/charts/rancher-gke-operator/1.1.100/Chart.yaml
+++ b/charts/rancher-gke-operator/1.1.100/Chart.yaml
@@ -5,4 +5,4 @@ home: https://github.com/rancher/gke-operator
 sources:
   - "https://github.com/rancher/gke-operator"
 version: 1.1.100
-appVersion: v1.1.1-rc6
+appVersion: v1.1.1-rc7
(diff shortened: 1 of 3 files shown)


thedadams (approved):

rmweir (approved):
LGTM

charts/rancher-gke-operator/1.1.100/Chart.yaml
```diff
@@ -5,4 +5,4 @@ home: https://github.com/rancher/gke-operator
 sources:
   - "https://github.com/rancher/gke-operator"
 version: 1.1.100
-appVersion: v1.1.1-rc6
+appVersion: v1.1.1-rc7
```
cbron, line 8: Since this is 1.1.1, this is the version for 2.6. Why do we still need this one in system-charts if we transitioned to feature charts ?

cmurphy: It's the forwardport of https://github.com/rancher/system-charts/pull/496. I thought that was the requirement?

cmurphy:
Not needed anymore, replacing with https://github.com/rancher/system-charts/pull/501

PR #500, opened 2021-07-08
[2.5] Fix gke-operator charts

cmurphy:
Sync 1.0.200 with 2.5.9 branch, move current 1.0.200 package to 1.0.300, bump gke-operator to latest for 1.0.300.

### DIFF
--- a/charts/rancher-gke-operator/1.0.300/questions.yaml
+++ b/charts/rancher-gke-operator/1.0.300/questions.yaml
@@ -1,0 +1,1 @@
+rancher_min_version: 2.5.8-patch1
--- a/charts/rancher-gke-operator/1.0.300/questions.yaml
+++ b/charts/rancher-gke-operator/1.0.300/questions.yaml
@@ -1,1 +1,1 @@
-rancher_min_version: 2.5.8-patch1
+rancher_min_version: 2.5.10-patch1
--- a/charts/rancher-gke-operator/1.0.200/questions.yaml
+++ b/charts/rancher-gke-operator/1.0.200/questions.yaml
@@ -1,1 +1,1 @@
-rancher_min_version: 2.5.8-patch1
+rancher_min_version: 2.5.9-rc5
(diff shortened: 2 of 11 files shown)


charts/rancher-gke-operator/1.0.200/Chart.yaml
```diff
@@ -5,4 +5,4 @@ home: https://github.com/rancher/gke-operator
 sources:
   - "https://github.com/rancher/gke-operator"
 version: 1.0.200
-appVersion: v1.1.1-rc5
+appVersion: v1.0.2-rc2
```
cbron: We'll need to remember to bump this to final release in both branches.

charts/rancher-gke-operator/1.0.300/questions.yaml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.5.8-patch1
```
cbron: Should be 2.5.10-patch1, as Rancher codebase won't support this version until then. We don't want rancher v2.5.9 eventually getting this from release-v2.5 when 2.5.10 goes out.

cbron (changes requested):
Minor change. Once 2.5.9 goes into release-v2.5 this may conflict, we'll need to pull 2.5.9 branch in here.

charts/rancher-gke-operator/1.0.300/questions.yaml
cmurphy: done

thedadams (dismissed):

thedadams (dismissed):

jiaqiluo:
@cmurphy  Can you update the version in https://github.com/rancher/rancher/issues/33528 once this PR is merged?

cbron (dismissed):
lgtm, note that the 1.0.200 min version changed.

charts/rancher-gke-operator/1.0.200/Chart.yaml
cmurphy: Will update after https://github.com/rancher/system-charts/pull/503

thedadams (approved):

cbron (approved):

PR #504, opened 2021-07-13
[dev-v2.6] Add missing rancher version constraints to chart questions

pennyscissors:
Issues: https://github.com/rancher/rancher/issues/33037, https://github.com/rancher/rancher/issues/35687
Needed to avoid exporting images from old charts.

There is an additional commit to add a workflow to validate questions which I thought could be helpful, but I can remove it if we don't want to be as strict with it.


.github/workflows/validate.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - dev-v2.6
+
+jobs:
+  validate-chart-questions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Validate all charts have a questions file
```
paynejacob: can we move this to scripts, developers should be able to run all validation locally as well.

paynejacob (changes requested):

.github/workflows/validate.yaml
paynejacob: talked offline, given the LOE to add done and current plans for this repo this makes sense to put here

paynejacob (dismissed):

.github/workflows/validate.yaml
pennyscissors: We discussed this offline and decided it is better to keep as is since system charts don't have a CI and it would be too much effort for little benefit.

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,29 @@
...
+              echo "$chart is missing a questions chart file"
+              exit 1
+            fi
+          done
+
+      - name: Validate all questions files have a `rancher_min_version` set
+        run: >
+          questions_files=$(find charts -type f -name questions.y\*ml)
+
+          grep -rL "rancher_min_version" $questions_files | sed 's/$/ is missing a rancher_min_version constraint/' | (! grep .)
```
aiyengar2: Can you add a comment about what this does? Also, this one-line could be rewritten to be easier to parse using if / else statements, e.g. saving the output of the first grep into a for loop and then running `echo` instead of `sed`.

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Validate all charts have a questions file
+        run: >
+          chart_dirs=$(find charts -type d -mindepth 2 -maxdepth 2)
+
+          for chart in $chart_dirs; do
+            if [[ ! -f $chart/questions.yaml ]] && [[ ! -f $chart/questions.yml ]]; then
+              echo "$chart is missing a questions chart file"
+              exit 1
+            fi
+          done
```
aiyengar2: wondering if both of these scripts should be located under `./scripts` instead. What are your thoughts?

aiyengar2 (commented):
Minor comments on the workflows, otherwise min version changes LGTM. Thanks for writing the CI!

.github/workflows/validate.yaml
aiyengar2: seems like this was the same concern brought up in https://github.com/rancher/system-charts/pull/504#discussion_r760588402. @paynejacob @PennyScissors if you both agree still, then I'm find with skipping this as well.

But I think we still need a comments for readability at least.

StrongMonkey (dismissed):

.github/workflows/validate.yaml
pennyscissors: I cleaned this up as suggested. I didn't add a comment because the name of the step and variable names should be enough to know what it does now.

pennyscissors:
> wondering if both of these scripts should be located under `./scripts` instead. What are your thoughts?

After discussing this with @aiyengar2, @StrongMonkey, and @cbron, it was decided we can remove the drone CI in favor of a Github Workflow to validate the questions files. The current drone CI is no-op, and looks as if it has been this way since the repo was created considering none of the branches have a Makefile and therefore it never runs. Since there won't be a scripts directory once the drone stuff is removed, and in reality a dev doesn't really need to validate the questions files locally, I will leave the logic in the workflow's yaml.

paynejacob (dismissed):

jiaqiluo (dismissed):
LGTM

jiaqiluo (approved):
LGTM. 

Talked with @PennyScissors  offline and got the following confirmation:
1. monitoring 0.1.5 does not have rancher_max_version since we need it for a use case related to k8s version compatibility
2. logging 0.3.0 does not have rancher_max_version since we need both it and the latest version cc @paynejacob 
3. for the rest of the charts, the rancher_max_version is set to 2.6.2 to minimize the influence.

PR #508, opened 2021-07-14
Sync gke-operator 1.0.200 package from 2.5.9

cmurphy:


cbron (approved):

cbron (changes requested):
This should be 1.0.300

thedadams (dismissed):

charts/rancher-gke-operator/1.0.300/values.yaml
```diff
@@ -4,7 +4,7 @@ global:
 gkeOperator:
   image:
     repository: rancher/gke-operator
-    tag: v1.1.1-rc7
+    tag: v1.0.3
 
 httpProxy: ""
 httpsProxy: ""
```
cbron, line 10: I think 1.0.300 should have more changes no ? Looks like questions.yaml is still set on old one: https://github.com/rancher/system-charts/blob/be12cf837e206bd3b8917b5cb3ea626f103427e0/charts/rancher-gke-operator/1.0.300/questions.yaml

cbron (approved):

thedadams (approved):

PR #514, opened 2021-08-06
25logging bump

paynejacob:


nickgerace (dismissed):

charts/rancher-logging/0.2.5/.helmignore
```diff
@@ -0,0 +1,21 @@
+# Patterns to ignore when building packages.
```
cbron, line 1: Is this a backport, or just the chart version change ? Needs some issue linked which explains what/why its changing.

charts/rancher-logging/0.2.5/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
cbron, line 1: We need to forwardport this chart to dev-v2.6 after it merges

charts/rancher-logging/0.2.5/Chart.yaml
```diff
@@ -0,0 +1,12 @@
...
+description: Rancher logging helm chart to support logging function in rancher
+icon: https://raw.githubusercontent.com/fluent/fluentd-docs/master/public/logo/Fluentd_square.png
+name: rancher-logging
+version: 0.2.5
```
cbron, line 5: Next time we should do the version bumps in the 2nd commit.

charts/rancher-logging/0.2.5/questions.yml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.3.4-rc1
```
cbron: I know we haven't updated this lately, but we should bump this to 2.5.10, so that earlier 2.5's stay stable.

charts/rancher-logging/0.2.5/charts/fluentd/charts/fluentd-linux/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+name: fluentd-linux
+version: 0.0.2
+appVersion: 1.6.3
```
cbron, line 3: Don't all of these sub-chart versions have to change too ? https://github.com/rancher/system-charts/blob/dev-v2.5/charts/rancher-logging/0.2.4/charts/fluentd/charts/fluentd-linux/Chart.yaml#L2-L3

cbron (changes requested):

charts/rancher-logging/0.2.5/charts/fluentd/charts/fluentd-linux/Chart.yaml
paynejacob: it looks like we don't usally upgrade them.  I imagine helm calls upgrade on them given they are packaged with the parent chart and they get upgraded.

paynejacob: As a helm best practice it should be updated because sub charts should be standalone.  Given we only use the sub charts here in this chart it doesn't affect anything.  This is something we may want to do going forward though.

cbron (approved):

PR #518, opened 2021-08-10
[2.6] Forward Port:  25logging bump

paynejacob:


cbron:
This one is just to ensure previously installed chart can be edited in newer version. 
original: https://github.com/rancher/system-charts/pull/514

cbron (approved):

PR #520, opened 2021-08-18
Changes for rancher-k3s-upgrader

kinarashah:
https://github.com/rancher/rancher/issues/34071

### DIFF
--- a/charts/rancher-k3s-upgrader/0.3.0/templates/configmap.yaml
+++ b/charts/rancher-k3s-upgrader/0.3.0/templates/configmap.yaml
@@ -8,7 +8,7 @@ data:
   SYSTEM_UPGRADE_CONTROLLER_THREADS: {{ .Values.systemUpgradeControllerThreads | default "2" | quote }}
   SYSTEM_UPGRADE_JOB_ACTIVE_DEADLINE_SECONDS: {{ .Values.systemUpgradeJobActiveDeadlineSeconds | default "900" | quote }}
   SYSTEM_UPGRADE_JOB_BACKOFF_LIMIT: {{ .Values.systemUpgradeJobBackoffLimit | default "99" | quote }}
-  SYSTEM_UPGRADE_JOB_IMAGE_PULL_POLICY: {{ .Values.systemUpgradeJobImagePullPolicy | default "Always" | quote }}
+  SYSTEM_UPGRADE_JOB_IMAGE_PULL_POLICY: {{ .Values.systemUpgradeJobImagePullPolicy | default "IfNotPresent" | quote }}
   SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE: {{ template "system_default_registry" . }}{{ .Values.kubectl.image.repository }}:{{ .Values.kubectl.image.tag }}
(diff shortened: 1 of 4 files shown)


charts/rancher-k3s-upgrader/0.3.0/templates/configmap.yaml
```diff
@@ -8,7 +8,7 @@ data:
   SYSTEM_UPGRADE_CONTROLLER_THREADS: {{ .Values.systemUpgradeControllerThreads | default "2" | quote }}
   SYSTEM_UPGRADE_JOB_ACTIVE_DEADLINE_SECONDS: {{ .Values.systemUpgradeJobActiveDeadlineSeconds | default "900" | quote }}
   SYSTEM_UPGRADE_JOB_BACKOFF_LIMIT: {{ .Values.systemUpgradeJobBackoffLimit | default "99" | quote }}
-  SYSTEM_UPGRADE_JOB_IMAGE_PULL_POLICY: {{ .Values.systemUpgradeJobImagePullPolicy | default "Always" | quote }}
+  SYSTEM_UPGRADE_JOB_IMAGE_PULL_POLICY: {{ .Values.systemUpgradeJobImagePullPolicy | default "IfNotPresent" | quote }}
```
nickgerace, line 11: Could this change have negative implications for upgrade scenarios, cleaned nodes, etc.?

kinarashah: Can't think of any, unless image is force pushed or something and tag remains the same. Any specific scenario standing out for you?

dramich: `IfNotPresent` is the usual default for a pull policy. If the image doesn't exist, then pull it, but always pulling breaks airgap setups. 

If you have a more specific example of where you think this could be a problem then we can answer more directly.

aiyengar2, line 11: Is there an implication to updating `IfNotPresent` here rather than specifying `systemUpgradeJobImagePullPolicy: IfNotPresent` in the values.yaml?

Seems like this changes a default that is expected in the SUC chart; could this break users in certain environments?

aiyengar2 (changes requested):

charts/rancher-k3s-upgrader/0.3.0/templates/configmap.yaml
aiyengar2: ah seems like Nick beat me to this in https://github.com/rancher/system-charts/pull/520/files#r691642993 and Dan responded. I will resolve this comment and approve!

aiyengar2 (dismissed):
LGTM

paynejacob (approved):

charts/rancher-k3s-upgrader/0.3.0/templates/configmap.yaml
nickgerace: I had originally thought that might need to pull again, but `IfNotPresent` covers this. Resolving

nickgerace (approved):

cbron (approved):

PR #530, opened 2021-09-02
Add securityPolicy to gke-operator

thedadams:
The securityContext is necessary when adding additional CAs.

Issue:


cmurphy (dismissed):
nit about the commit message, it's `securityContext`

cmurphy:
This doesn't bring the chart into alignment with what's in gke-operator currently, do we need to add this too https://github.com/rancher/gke-operator/blob/6b98808efcf8067480818eecca2ceecfde242608/charts/gke-operator/templates/deployment.yaml#L39-L44 ?

thedadams:
@cmurphy Good catches! Updated.

cmurphy (approved):

jiaqiluo (approved):
LGTM

charts/rancher-gke-operator/1.1.200/Chart.yaml
```diff
@@ -4,5 +4,5 @@ description: A Helm chart for provisioning GKE clusters
 home: https://github.com/rancher/gke-operator
 sources:
   - "https://github.com/rancher/gke-operator"
-version: 1.0.400
```
jiaqiluo, line 7: 1.0.400 is an unreleased version so we can overwrite the version.

charts/rancher-gke-operator/1.1.200/Chart.yaml
```diff
@@ -4,5 +4,5 @@ description: A Helm chart for provisioning GKE clusters
...
-appVersion: v1.1.1-rc10
+version: 1.1.200
```
cbron, line 7: Needs to be updated here: https://github.com/rancher/rancher/issues/33528

cbron (approved):
Lgtm. We should dev-validate that the basics here are working before sending to qa, meaning that the chart sets cert appropriately.

charts/rancher-gke-operator/1.1.200/Chart.yaml
thedadams: Updated version in linked issue.

PR #533, opened 2021-09-09
fix helm install failure 

jiaqiluo:
issue: https://github.com/rancher/rancher/issues/33831

Problem:
The current design is to skip the creation of all CRDs if the API group version `monitoring.coreos.com/v1` exists in the cluster.

If one of the crd, for example, the crd `servicemonitors.monitoring.coreos.com/v1` does not exist, Helm install will fail and report the following error:

```
Failed to install app cluster-alerting. Error: validation failed: unable to recognize "": no matches for kind "ServiceMonitor" in the API group version "monitoring.coreos.com/v1"
```

Fix:

Instead of checking the API group, we now check each individual kind and create the missing one while installing the app.  ( **see the 3rd commit for the fix**)


charts/rancher-monitoring/v0.3.1/Chart.yaml
```diff
@@ -7,8 +7,8 @@ maintainers:
 name: rancher-monitoring
 sources:
 - https://github.com/coreos/prometheus-operator
-version: 0.3.0
-appVersion: 0.3.0
+version: 0.3.1
```
aiyengar2, line 10: Need to update question.yml as well

aiyengar2 (changes requested):

charts/rancher-monitoring/v0.3.1/Chart.yaml
jiaqiluo: fixed!

brendarearden (dismissed):
LGTM

aiyengar2 (dismissed):
LGTM

charts/rancher-monitoring/v0.3.1/templates/crds.yaml
```diff
@@ -18,7 +18,9 @@ spec:
     plural: alertmanagers
   scope: Namespaced
   version: v1
+{{- end }}
 ---
```
cbron, line 21: These should be included in the if blocks, or you'll get: 
```
---
---
---
```

charts/rancher-monitoring/v0.3.1/questions.yml
```diff
@@ -0,0 +1 @@
+rancher_min_version: 2.6.1
```
cbron: The 0.3.0 chart should have a max of 2.6.0

charts/rancher-monitoring/v0.3.1/templates/crds.yaml
jiaqiluo: fixed.

charts/rancher-monitoring/v0.3.1/questions.yml
jiaqiluo: `rancher_max_version: 2.6.0` is added to monitoring 0.3.0

cbron (approved):

PR #537, opened 2021-09-27
Add rancher-k3s-upgrade 0.3.1

thedadams:
Related issues:
https://github.com/rancher/rancher/issues/34908
https://github.com/rancher/rancher/issues/34908

PR #540, opened 2021-10-05
Add rancher-k3s-upgrade 0.3.1

thedadams:
Issues:
https://github.com/rancher/rancher/issues/34908
https://github.com/rancher/rancher/issues/32633

This is the same as the reverted PR https://github.com/rancher/system-charts/pull/537

### DIFF
--- a/charts/rancher-k3s-upgrader/0.3.1/values.yaml
+++ b/charts/rancher-k3s-upgrader/0.3.1/values.yaml
@@ -1,0 +1,12 @@
+global:
+  systemDefaultRegistry: ""
+
+systemUpgradeController:
+  image:
+    repository: rancher/system-upgrade-controller
+    tag: v0.7.5
+
+kubectl:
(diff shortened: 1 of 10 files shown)


charts/rancher-k3s-upgrader/0.3.1/values.yaml
```diff
@@ -0,0 +1,12 @@
+global:
+  systemDefaultRegistry: ""
+
+systemUpgradeController:
+  image:
+    repository: rancher/system-upgrade-controller
+    tag: v0.8.0-rc.1
```
kinarashah: v0.8.0 has been tagged now, any reason we still want to use the rc?

thedadams: No, I didn't now the new version had been tagged. Updated.

kinarashah (approved):

PR #542, opened 2021-10-06
Downgrade gke-operator to 1.1.1

thedadams:


charts/rancher-gke-operator/1.1.100/Chart.yaml
```diff
@@ -4,5 +4,5 @@ description: A Helm chart for provisioning GKE clusters
 home: https://github.com/rancher/gke-operator
 sources:
   - "https://github.com/rancher/gke-operator"
-version: 1.1.200
-appVersion: v1.1.2-rc1
+version: 1.1.100
```
cbron, line 7: Needs updating here: https://github.com/rancher/rancher/issues/33528
Otherwise this change is fine, 1.1.x never was released.

cbron (approved):

cmurphy (approved):

PR #544, opened 2021-10-12
Rebase on release

thedadams:
There was an issue with gke-operator commit history that made it look like 1.0.300 was a new charts going into v2.5.10. This is a rebase to line up the commit histories to fix this issue.


charts/rancher-eks-operator/1.0.1000/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+name: rancher-eks-operator
+description: A Helm chart for provisioning EKS clusters
+home: https://github.com/rancher/eks-operator
+sources:
+  - "https://github.com/rancher/eks-operator"
+version: 1.0.1000
```
jiaqiluo, line 7: I am confused here:
this is a new version that does not exist in the released-v2.5 branch https://github.com/rancher/system-charts/tree/release-v2.5/charts/rancher-eks-operator
How could it show up as an addition in this PR? 

Is my understanding right that your PR is to pull the release-v2.5 into the dev-v2.5 branch?

thedadams: The same thing is happening for gke-operator 1.1.100

This is my understanding:
The files are not new, but the commits that contain the files are new. This is the same thing that was happening in the `dev-2.5` to `release-v2.5` PR, which led to this rebase. After this rebase, the release PR should have a cleaner history.

thedadams: This is what the `dev-v2.5` to `release-v2.5` PR would look like after this is merged: https://github.com/rancher/system-charts/compare/release-v2.5...thedadams:rebase-on-release
You can see that the eks-operator files are still listed as new (which they should be) and only one new version of gke-operator is listed (1.1.100) which is also correct.

jiaqiluo: @thedadams  the above dev-to-release mock PR looks good.

jiaqiluo (approved):
LGTM
As @thedadams  shows, this PR should make dev-2.5 be built at the top of release-v2.5

 > This is what the `dev-v2.5` to `release-v2.5` PR would look like after this is merged: [release-v2.5...thedadams:rebase-on-release](https://github.com/rancher/system-charts/compare/release-v2.5...thedadams:rebase-on-release)
> You can see that the eks-operator files are still listed as new (which they should be) and only one new version of gke-operator is listed (1.1.100) which is also correct.

cmurphy (approved):

cbron (approved):
Hard to tell whats really going on here, I believe all of these commits are just ones that are newer than the 2.5.9 PR merge, and inserting those 2.5.9 commits before them changes them. Seems fairly low risk to merge, since all these changes are just the current state of dev-v2.5 as far as I can tell.

PR #541, opened 2021-10-06
Release System Charts for 2.5.10

thedadams:
https://github.com/rancher/rancher/issues/33528

PR #545, opened 2021-10-13
Standardize min/max rancher versions for new charts

thedadams:
In 2.5.10, there are three charts that are introduced. The min
versions for these charts are inconsistent. This change fixes
these inconsistencies.


aiyengar2:
addresses https://github.com/rancher/system-charts/pull/541#pullrequestreview-778816506

aiyengar2 (dismissed):
LGTM. Waiting on @cbron to confirm whether the rancher_max_versions are necessary here, not sure what our policy is for patch releases.

cbron (approved):

PR #546, opened 2021-10-15
Move logging 0.2.5 released in 2.5.10 to 2.6

thedadams:


charts/rancher-logging/0.2.5/questions.yml
```diff
@@ -0,0 +1,2 @@
+rancher_min_version: 2.5.10-alpha1
+rancher_max_version: 2.5.99-0
```
jiaqiluo, line 2: nice.

jiaqiluo (changes requested):
the rancher-logging chart LGTM. 

How about the eks-operator and aks-operator? Although they were moved to rancher/charts, we still need to forward-port those two new versions, right?

charts/rancher-logging/0.2.5/questions.yml
thedadams: No, we don't need the eks and gke operator charts. When Rancher 2.6+ starts up, it will uninstall the existing eks/gke-operator charts and install the rancher/charts version.

There are currently no eks/gke-operator charts in system-charts in 2.6

jiaqiluo: Oh right! I forgot that they will be uninstalled. 😅

jiaqiluo (dismissed):

aiyengar2 (dismissed):
LGTM, based on https://github.com/rancher/rancher/issues/33528, the only other charts were operator charts that should not need to be forward-ported since they are auto-upgraded on Rancher upgrades.

thedadams:
https://github.com/rancher/rancher/issues/32842

cbron:
Switched to release branch

cbron (approved):
lgtm, will wait for others than merge, and we can pull back into dev

jiaqiluo (approved):
LGTM.  
Since the PR goes to the release-v2.6 directly now, we will need to pull the release-v2.6 back into to dev-v2.6.

aiyengar2 (approved):

PR #547, opened 2021-10-20
Rebase on release-v2.6 after 2.5.10 release

thedadams:
[This PR](https://github.com/rancher/system-charts/pull/546) forward-ported rancher-logging which was released in 2.5.10. In order to make the next release of 2.6.x go smoothly, dev-v2.6 needs to rebased on release-v2.6. This PR does exactly that.


charts/rancher-k3s-upgrader/0.3.1/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+name: rancher-k3s-upgrader
+description: Enables a k3s or rke2 cluster to update itself by reacting to Plan CRs.
+  Users do not need to manually upgrade this app. It will be automatically upgraded to the latest version when upgrading a cluster.
+home: https://github.com/rancher/system-charts/blob/dev-v2.6/charts/rancher-k3s-upgrader
+sources:
+  - "https://github.com/rancher/system-charts/blob/dev-v2.6/charts/rancher-k3s-upgrader"
+version: 0.3.1
```
jiaqiluo, line 8: It looks like this chart already exists in the dev-v2.6 branch, why does it show in this PR as an addition again?
https://github.com/rancher/system-charts/tree/dev-v2.6/charts/rancher-k3s-upgrader/0.3.1

jiaqiluo (changes requested):

charts/rancher-k3s-upgrader/0.3.1/Chart.yaml
thedadams: It shows up because this is a rebase. This was the same problem we had previously in this PR: https://github.com/rancher/system-charts/pull/544

By doing this rebase now, we will avoid this problem when the next `dev-v2.6` -> `release-v2.6` PR happens.

jiaqiluo (approved):
LGTM

charts/rancher-k3s-upgrader/0.3.1/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
```
cbron, line 1: This is just a chart that was merged to dev-v2.6 since we updated release branch ?

thedadams: Yes, that's correct.

cbron (approved):

PR #549, opened 2021-11-16
adding logging v0.3.1with windows ltsc2022 support

luthermonson:
rancher/rancher#36446


charts/rancher-logging/0.3.1/charts/fluentd/charts/fluentd-windows/values.yaml
```diff
@@ -0,0 +1,52 @@
+nameOverride: fluentd
+labels: {}
+
+image:
+  os: windows
+  repository: rancher/fluentd
+  tag: v0.1.31
```
jiaqiluo, line 7: The tag is the only change in this new chart version compared to the previous version. 
In the previous version, the tag is `v0.1.30`

jiaqiluo (approved):

PR #552, opened 2021-11-22
fix rancher logging on v1.22

paynejacob:
https://github.com/rancher/rancher/issues/35620

# Problem
The rancher logging controller depends on finding "a latest version" of the chart existing.  If the chart does not exist we the controller will return an error even if logging is not enabled.

# Solution
Give rancher logging an empty chart in v1.22 to deploy.   This will have the same effect as uninstalling rancher logging if a cluster is upgraded to v1.22 and suppresses the errors when

### DIFF
--- a/charts/rancher-logging/0.4.0/.helmignore
+++ b/charts/rancher-logging/0.4.0/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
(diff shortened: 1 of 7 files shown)


kinarashah (approved):

nickgerace:
Shouldn't we copy the old chart in the first commit to see the diff?

paynejacob:
@nickgerace I can redo the commits if you want but I opted to do 1 commit here because I just delete the templates anyway so it would be difficult to review.

nickgerace (approved):

PR #577, opened 2022-11-14
Bring back changes from release-v2.6 to dev-v2.6

aiyengar2:


prachidamle:
These are 3 "git merge" commits that were present in release/v2.6 and were not back ported to dev-v2.6. We need to get these in to correct the commit history between dev-v2.6 and release-v2.6

prachidamle (approved):

PR #581, opened 2023-02-07
Conditionally enable psp

HarrisonWAffel:
## Problem: 
The rancher-k3s-upgrader chart will deploy PSP's if the kubernetes version running on the downstream cluster still has the capability to do so. This presents a problem when upgrading a cluster to a version which no longer has PSP support. Helm will be unable to upgrade or uninstall the upgrader due to it holding references to PSP's which can now no longer be removed.

## Solution: 
Add a new field to the rancher-k3s-upgrader chart, `psp: enabled` and set its default value to `false`. This value will be used within Rancher to determine if PSP's are enabled or disabled on any version of Kubernetes, and will be used to block upgrades to 1.25+ until a new version of the upgrader is deployed which has psp's disabled. 

I've added a new version, `0.4.1`, though I don't believe `0.4.0` has been released yet. If it's better that I modify that version let me know.

### DIFF
--- a/charts/rancher-k3s-upgrader/0.4.0/templates/validate-psp-install.yaml
+++ b/charts/rancher-k3s-upgrader/0.4.0/templates/validate-psp-install.yaml
@@ -1,0 +1,7 @@
+#{{- if gt (len (lookup "rbac.authorization.k8s.io/v1" "ClusterRole" "" "")) 0 -}}
+#{{- if .Values.global.cattle.psp.enabled }}
+#{{- if not (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}
+#{{- fail "The target cluster does not have the PodSecurityPolicy API resource. Please disable PSPs in this chart before proceeding." -}}
+#{{- end }}
+#{{- end }}
+#{{- end }}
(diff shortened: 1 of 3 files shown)


jiaqiluo (changes requested):
overall LGTM.
As to the version, we can use 0.4.0 which is unreleased.

HarrisonWAffel:
I've updated this PR to change the contents of `0.4.0` instead of introducing `0.4.1`

jiaqiluo (dismissed):

snasovich (dismissed):
LGTM, some pending discussions on what `psp.enabled` default should be.

charts/rancher-k3s-upgrader/0.4.0/templates/validate-psp-install.yaml
```diff
@@ -0,0 +1,7 @@
+#{{- if gt (len (lookup "rbac.authorization.k8s.io/v1" "ClusterRole" "" "")) 0 -}}
```
jiaqiluo, line 1: Do you mean to check if there is any ClusterRole or if the particular one named `system-upgrade-controller-psp` is in the cluster?   if the answer is the latter, we should specify the namespace and name.

HarrisonWAffel: This is the standard validation file that's now being used across all charts, from what I understand its required to ensure that `helm template` will still work with these conditions added

jiaqiluo: OK. Though what it checks seems useless to me:  check if there is any `ClusterRole.rbac.authorization.k8s.io/v1` object in the cluster, it does not hurt at least 😆

jiaqiluo (approved):

snasovich (approved):

PR #590, opened 2023-05-24
fix: Allow rancher-monitoring to be installed on k8s clusters above v2

alexlokshin-czi:
When using rancher terraform provider with `enable_cluster_monitoring = true` to import clusters 1.22 and above, clusters are imported with an error: `Template system-library-rancher-monitoring incompatible with rancher version or cluster's [*****] kubernetes version`; example of a resource block is below:
```
resource "rancher2_cluster" "cluster" {
  name                      = "cluster"
  enable_cluster_monitoring = true
}
```


rohitsakala:
@alexlokshin-czi how do use rancher monitoring in rancher 2.7 version. Can you please tell me the steps ? Thanks

PR #594, opened 2023-08-28
Create next version of rancher-k3s-upgrader chart

thatmidwesterncoder:
Tracking issue: https://github.com/rancher/rancher/issues/41791

Creating a new chart and bumping the system-upgrade-controller to v0.13.1.

PR #601, opened 2024-01-19
[2.7.11] Support k8s 1.27 for system-upgrade-controller

thatmidwesterncoder:
Issue: https://github.com/rancher/rancher/issues/44114

Pulling the changes to support k8s v1.27 from https://github.com/rancher/system-charts/pull/594

### DIFF
--- a/charts/rancher-k3s-upgrader/0.6.0/Chart.yaml
+++ b/charts/rancher-k3s-upgrader/0.6.0/Chart.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+name: rancher-k3s-upgrader
+description: Enables a k3s or rke2 cluster to update itself by reacting to Plan CRs.
+  Users do not need to manually upgrade this app. It will be automatically upgraded to the latest version when upgrading a cluster.
+home: https://github.com/rancher/system-charts/blob/dev-v2.7/charts/rancher-k3s-upgrader
+sources:
+  - "https://github.com/rancher/system-charts/blob/dev-v2.7/charts/rancher-k3s-upgrader"
+version: 0.5.0
+appVersion: v0.11.0
+kubeVersion: '>= 1.23.0-0'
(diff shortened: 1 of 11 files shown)


HarrisonWAffel (approved):

charts/rancher-k3s-upgrader/0.6.0/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+name: rancher-k3s-upgrader
+description: Enables a k3s or rke2 cluster to update itself by reacting to Plan CRs.
+  Users do not need to manually upgrade this app. It will be automatically upgraded to the latest version when upgrading a cluster.
+home: https://github.com/rancher/system-charts/blob/dev-v2.7.11/charts/rancher-k3s-upgrader
+sources:
+  - "https://github.com/rancher/system-charts/blob/dev-v2.7.11/charts/rancher-k3s-upgrader"
```
jiaqiluo, line 7: Nit:  What will be the ultimate branch for hosting this chart, will it be `release-v2.7`? 
How did we handle it in older versions of this chart?

jiaqiluo (changes requested):

jiaqiluo (approved):
I think my above question should not be a blocker for merging the PR.

charts/rancher-k3s-upgrader/0.6.0/Chart.yaml
thatmidwesterncoder: So the chart version already exists in `dev-v2.8` and above but with different rancher requirements - will the 2.7.11 branch get merged back into 2.7? If so I'm happy to update it to just `release-v2.7`.

jiaqiluo: I think about it again, I think the branch `dev-v2.7.11` will be there forever as a reference to the source, so it should be ok to keep the file as it is.

PR #609, opened 2024-08-27
Introduce v0.7.0, utilizing SUC v0.13.4

HarrisonWAffel:
This PR introduces v0.7.0 of the rancher-k3s-upgrader, which introduces support for SUC v0.13.4. v0.13.4 is needed so that Rancher can configure the security context for SUC plans, which is required for the deployment of host process pods on Windows nodes. Without this updated chart, Rancher will not be able to deploy plans to Windows node on Rancher 2.8.x. 

No other aspects of the chart have been modified, but let me know if values such as the minimum kubeVersion should be bumped as well.

### DIFF
--- a/charts/rancher-k3s-upgrader/0.7.0/Chart.yaml
+++ b/charts/rancher-k3s-upgrader/0.7.0/Chart.yaml
@@ -5,6 +5,6 @@ description: Enables a k3s or rke2 cluster to update itself by reacting to Plan 
 home: https://github.com/rancher/system-charts/blob/dev-v2.8/charts/rancher-k3s-upgrader
 sources:
   - "https://github.com/rancher/system-charts/blob/dev-v2.8/charts/rancher-k3s-upgrader"
-version: 0.6.0
-appVersion: v0.13.1
+version: 0.7.0
+appVersion: v0.13.4
 kubeVersion: '>= 1.23.0-0'
--- a/charts/rancher-k3s-upgrader/0.7.0/Chart.yaml
+++ b/charts/rancher-k3s-upgrader/0.7.0/Chart.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+name: rancher-k3s-upgrader
+description: Enables a k3s or rke2 cluster to update itself by reacting to Plan CRs.
+  Users do not need to manually upgrade this app. It will be automatically upgraded to the latest version when upgrading a cluster.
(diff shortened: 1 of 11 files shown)


charts/rancher-k3s-upgrader/0.6.0/Chart.yaml
```diff
@@ -6,5 +6,5 @@ home: https://github.com/rancher/system-charts/blob/dev-v2.8/charts/rancher-k3s-
 sources:
   - "https://github.com/rancher/system-charts/blob/dev-v2.8/charts/rancher-k3s-upgrader"
 version: 0.6.0
-appVersion: v0.13.1
+appVersion: v0.13.4
```
Oats87: is this a mistake? I don't think we should mutate the `0.6.0` chart

charts/rancher-k3s-upgrader/0.7.0/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+name: rancher-k3s-upgrader
+description: Enables a k3s or rke2 cluster to update itself by reacting to Plan CRs.
+  Users do not need to manually upgrade this app. It will be automatically upgraded to the latest version when upgrading a cluster.
+home: https://github.com/rancher/system-charts/blob/dev-v2.8/charts/rancher-k3s-upgrader
+sources:
+  - "https://github.com/rancher/system-charts/blob/dev-v2.8/charts/rancher-k3s-upgrader"
+version: 0.6.0
```
Oats87: version should be `0.7.0` right?

Oats87 (changes requested):

charts/rancher-k3s-upgrader/0.6.0/Chart.yaml
HarrisonWAffel: Good catch, split this into two commits and must have updated the wrong directory

charts/rancher-k3s-upgrader/0.7.0/Chart.yaml
HarrisonWAffel: Updated, thanks!

Oats87 (approved):

jiaqiluo (approved):
LGTM.  
Just to confirm there will be a PR to forward port the new chart to the dev-v2.9 branch, right?

PR #612, opened 2024-09-30
[dev-v2.8] Remove logging and monitoring charts

mallardduck:
Per recent slack chats, these charts are not necessary for use anymore as all logging and monitoring charts should come from `rancher/charts` repo for 2.7+; doing a manual verification I cannot find references to indicate that these should be kept around.

If anyone happens to incorrectly still be using these versions then they likely have larger issues than removing this will cause and should opt to update to a version coming from `rancher/charts` sources. We can likely also backport this to 2.7 as well - though that is less urgent/necessary as that version will go EOL sooner than later.


ericpromislow (approved):
Same comment for 613:

There's a lot of changes here, but if I do a diff nothing else is deleted:
```
$ git diff dev-v2.7 | grep '^diff'  | grep -v -e 'charts/rancher-logging' -e 'charts/rancher-monitoring' | wc
       0       0       0
```

Note that for whatever reason the GHA workflow selects only v2.6 and 2.7

PR #613, opened 2024-09-30
[dev-v2.7] Remove logging and monitoring charts

mallardduck:
A backport of https://github.com/rancher/system-charts/pull/612


ericpromislow (approved):
There's a lot of changes here, but if I do a diff nothing else is deleted:
```
$ git diff dev-v2.7 | grep '^diff'  | grep -v -e 'charts/rancher-logging' -e 'charts/rancher-monitoring' | wc
       0       0       0
```

Note that for whatever reason the GHA workflow selects only v2.6 and 2.7

PR #614, opened 2024-09-30
[release-v2.7] Remove logging and monitoring charts

mallardduck:
A `release-` targeting version of: https://github.com/rancher/system-charts/pull/613

RE:

> Per recent slack chats, these charts are not necessary for use anymore as all logging and monitoring charts should come from rancher/charts repo for 2.7+; doing a manual verification I cannot find references to indicate that these should be kept around.
>
> If anyone happens to incorrectly still be using these versions then they likely have larger issues than removing this will cause and should opt to update to a version coming from rancher/charts sources. We can likely also backport this to 2.7 as well - though that is less urgent/necessary as that version will go EOL sooner than later.


nicholasSUSE (approved):
Assuming these charts are not being used by anyone, as per the PR owner. 
LGTM.

joshmeranda (approved):

jbiers (approved):

